PR #1, opened 2018-11-10
Initial Signature trait and Error enum

tarcieri:
Adds a `Signature` trait which serialized signature types implement.

Adds an initial `Error` type to be returned from `Signature::from_bytes` in the event the signature is not valid.

### DIFF
--- a/signature-crate/src/error.rs
+++ b/signature-crate/src/error.rs
@@ -1,0 +1,25 @@
+//! Signature error types
+
+use core::fmt;
+
+/// Errors which can occur during signature operations
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum Error {
+    /// Signature failed to verify
+    SignatureInvalid,
+}
+
+impl Error {
+    /// Get string description of this error
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Error::SignatureInvalid => "signature verification failed",
+        }
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
--- a/signature-crate/src/lib.rs
+++ b/signature-crate/src/lib.rs
@@ -1,1 +1,22 @@
+//! Digital signature types and traits
+
 #![crate_name = "signature"]
+#![crate_type = "lib"]
+#![no_std]
+#![cfg_attr(
+    all(feature = "nightly", not(feature = "std")),
+    feature(alloc)
+)]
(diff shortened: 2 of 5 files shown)


signature-crate/src/error.rs
```diff
@@ -0,0 +1,25 @@
+//! Signature error types
+
+use core::fmt;
+
+/// Errors which can occur during signature operations
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub enum Error {
+    /// Signature failed to verify
+    SignatureInvalid,
```
newpavlov: Are you sure about having single enum instead of several ZST types? [Here](https://github.com/briansmith/ring/blob/master/src/error.rs#L23) is for example a `ring`'s take on this.

Also it's probably better to implement `std::error::Error` trait behind `std` feature flag.

signature-crate/src/lib.rs
```diff
@@ -1 +1,22 @@
+//! Digital signature types and traits
+
 #![crate_name = "signature"]
+#![crate_type = "lib"]
```
newpavlov: Is there a reason for `crate_name` and `crate_type` attributes? I see it sometimes, but I don't know about why are they used.

You also can add `#![doc(html_logo_url ="https://raw.githubusercontent.com/RustCrypto/meta/master/logo_small.png")]` for the nice logo in docs. :)

signature-crate/src/error.rs
tarcieri: Yes, actually I just wrote up my thoughts about that here:

https://www.reddit.com/r/rust/comments/9wml68/introducing_mundane_a_new_cryptography_library/e9ma3fe/

While what *ring* does is great for *ring*, these traits are intended to also accommodate hardware-backed signers, and therefore need richer error types to express device failures, and are also fallible in ways software implementations are not. 

I can see how a single-member `enum` looks like YAGNI but believe me, I will need it 😉 

There's presently no `std::error::Error` impl, although I can certainly add one upstream in Signatory.

signature-crate/src/lib.rs
tarcieri: Perhaps this is pointless boilerplate. Also yes I know about `html_logo_url` 😉

signature-crate/src/error.rs
newpavlov: Hm, I wonder if it will be sensible to make a signature verification error an associated type. Unfortunately we don't have stable associated type defaults yet, so it will make implementing this trait a bit more verbose.

tarcieri: One of my personal goals is trait object safety, which precludes something like that.

PR #2, opened 2018-11-10
.travis.yml: Initial Travis CI configuration

tarcieri:

PR #3, opened 2019-03-24
.travis.yml: Check rustfmt, clippy, docs build, WASM + no_std target builds

tarcieri:
Some improvements in response to https://github.com/RustCrypto/signatures/pull/2#pullrequestreview-174880373

Doesn't quite do as an elaborate `no_std` check as the linked example there, however this is at least an improvement over not checking those targets whatsoever.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,46 @@ language: rust
 cache: cargo
 
 rust:
+  - 1.31.0
   - stable
   - beta
   - nightly
 
+install:
+  - rustup component add rustfmt
+  - rustup component add clippy
+  - rustup target add thumbv7em-none-eabihf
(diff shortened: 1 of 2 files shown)


tarcieri:
@newpavlov would appreciate some retroactive feedback, but otherwise I think this is at least an improvement form what it was previously

newpavlov:
In my previous experience with `rustfmt` I think it's better to allow failure for it. It results in less churn for outside contributors and protects against potential default formatting changes in `rustfmt`. But otherwise looks good! I probably should add something similar to other repositories as well.

tarcieri:
I'm personally a big `rustfmt` fan but yeah, I can add it to allowed failures...

PR #5, opened 2019-03-24
Improved signature::error::Error with cause

tarcieri:
Replaces the previous error `enum` with a struct that, on `std`, contains an optional cause which is captured as `Box<std::error::Error>`, but is otherwise opaque.

This allows the core signature traits to remain object safe (since they don't need a generic error type) but also allows passing through a provider-specific error type, e.g. for HSMs which may have any number of errors which prevent a signature operation from occuring (e.g. I/O errors, authentication errors, invalid key identifiers).

It also includes downcast methods that conveniently allow extracting provider-specific errors as concrete types if so desired.

### DIFF
--- a/signature-crate/src/error.rs
+++ b/signature-crate/src/error.rs
@@ -1,25 +1,86 @@
 //! Signature error types
 
-use core::fmt;
+use core::fmt::{self, Display};
 
-/// Errors which can occur during signature operations
-#[derive(Copy, Clone, Eq, PartialEq, Debug)]
-pub enum Error {
-    /// Signature failed to verify
-    SignatureInvalid,
+#[cfg(feature = "std")]
+use std::{boxed::Box, error::Error as StdError};
+
+/// Signature errors
+#[derive(Debug, Default)]
+pub struct Error {
+    /// Cause of the error (if applicable)
+    #[cfg(feature = "std")]
+    cause: Option<Box<dyn StdError>>,
 }
 
 impl Error {
-    /// Get string description of this error
-    pub fn as_str(self) -> &'static str {
-        match self {
-            Error::SignatureInvalid => "signature verification failed",
+    /// Create a new error with no cause
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Create a new error from a cause
+    #[cfg(feature = "std")]
+    pub fn from_cause<E>(cause: E) -> Self
+    where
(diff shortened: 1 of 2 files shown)


signature-crate/src/error.rs
```diff
@@ -1,25 +1,86 @@
...
-    SignatureInvalid,
+#[cfg(feature = "std")]
+use std::{boxed::Box, error::Error as StdError};
+
+/// Signature errors
+#[derive(Debug, Default)]
+pub struct Error {
+    /// Cause of the error (if applicable)
+    #[cfg(feature = "std")]
+    cause: Option<Box<dyn StdError>>,
```
tarcieri: An alternative to using `std::error::Error` is the [failure](https://github.com/rust-lang-nursery/failure) crate. However, while I'm a fan of this crate, I'm not sure it buys us anything here, or if it makes sense for all downstream crates to use failure.

newpavlov: IIRC `failure` is compatible with the `Error` trait, so I think it's better to leave it as-is. Backtraces will become less detailed, but I think it's fine.

tarcieri:
@newpavlov does this look good to you otherwise?

newpavlov:
I may be worth to look at `rand` error type for reference. It has a somewhat similar field, but it created some [isssues](https://github.com/rust-random/rand/issues/738). In `getrandom` we have introduced a different [error type](https://github.com/rust-random/getrandom/blob/master/src/error.rs#L21), which is essentially an error code. But I am not sure if such approach is applicable here.

tarcieri:
Yeah, it's a tricky set of constraints to attempt to align. I know `failure` has a similarly weird split in its internals for similar reasons.

I would like to hope that doesn't actually manifest as a problem in practice, because I think all of the crates which will want to impl traits from this crate which also support `no_std` will want to use the opaque error type. I think most crates that talk to HSMs will be dependent on `std`, however I can see myself eventually making something like my `yubihsm` crate `no_std` friendly.

Really the most important thing to me is having a way to at least pass through an error message describing what happened, in which case a `&'static str` is an alternative. However, I think it'd be really nice to be able to pass through and downcast the whole error object, especially for things like generating exception reports containing backtraces for HSM provider errors.

tarcieri:
I'm going to go with: this is an improvement over what was there before, and try to keep making progress.

PR #9, opened 2019-03-26
Add UseDigestTo* marker trait and blanket Sign/Verify impls

tarcieri:
This commit makes the digest for `SignDigest` and `VerifyDigest` an associated type rather than a generic, and also adds two marker traits:

- `UseDigestToSign`: opt into blanket impl of `Sign` for `SignDigest`
- `UseDigestToVerify`: opt into blanket impl of `Verify` for `VerifyDigest`

The reason for requiring a marker trait is there are cases where the same keys can be used with an IUF mode of an algorithm which is distinct from the non-IUF form, e.g. Ed25519 vs Ed25519ph. In this case, the underlying keys are the same, but the signature algorithm when using the
IUF mode is distinct.

### DIFF
--- a/signature-crate/src/lib.rs
+++ b/signature-crate/src/lib.rs
@@ -27,4 +27,4 @@ pub mod sign;
 mod signature;
 pub mod verify;
 
-pub use crate::{error::Error, sign::Sign, signature::Signature, verify::Verify};
+pub use crate::{error::Error, sign::*, signature::*, verify::*};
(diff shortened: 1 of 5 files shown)


tarcieri:
One drawback of this approach is it means `SignDigest` and `VerifyDigest` are no longer object safe.

To preserve that, one option is to move the associated type to the marker trait. It feels a bit redundant, but I'll go ahead and try that and push a second commit so we can compare.

tarcieri:
@newpavlov I have a new idea about how to do this which I think is a little more in the middle between our approaches and better describes the underlying concepts.

Will open as a new PR.

PR #11, opened 2019-03-29
Rename to Signer/Verifier; add 'digest' module

tarcieri:
Some sweeping refactoring here which hopefully cleans up the project organization and helps simplify its implementation:

- `Sign` => `Signer`, `Verify` => `Verifier`
- Move all `digest`-related functionality under the `digest` module:
  - `signature::digest::Signer`
  - `signature::digest::Verifier`
- Re-export `::digest::Digest` as `signature::digest::Digest`
- Remove `Send` + `Sync` bounds on traits: if people wants these bounds
  they can add them themselves.

### DIFF
--- a/signature-crate/src/digest/mod.rs
+++ b/signature-crate/src/digest/mod.rs
@@ -1,0 +1,11 @@
+//! Support for using hash functions that impl the `Digest` trait in order
+//! to hash the input message in order to compute a signature.
+
+mod signer;
+mod verifier;
+
+/// Re-export the `Digest` trait from the `digest` crate, as it's the main
+/// trait this module depends on.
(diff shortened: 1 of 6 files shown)


tarcieri:
@newpavlov many of these things were your suggestions, so hopefully aside from the `digest` module stuff this isn't to controversial.

That said I'd like to go ahead and merge this and immediately open another PR to address the blanket impl stuff we've been trying to do in #9 and #10, so if you do have feedback hopefully I can address it there (or perhaps in a separate PR).

tarcieri:
Okay, #12 is up, and contains the changes these were a setup for. Hopefully that helps illustrate how simple it actually is, versus if all of these were done in the same PR.

PR #12, opened 2019-03-29
Digestable signatures

tarcieri:
Adds a marker trait for `Signature` types computable as `S(H(m))` where:

- `S`: signature algorithm
- `H`: hash (a.k.a. digest) function
- `m`: message

For signature types that implement this trait, a blanket impl of `Signer` will be provided for all types that `impl signature::digest::Signer`.

### DIFF
--- a/signature-crate/src/digest/digestable.rs
+++ b/signature-crate/src/digest/digestable.rs
@@ -25,7 +25,7 @@ where
     T: digest::Signer<S::Digest, S>,
 {
     fn sign(&self, msg: &[u8]) -> Result<S, Error> {
-        self.sign_digest(S::Digest::new().chain(msg))
+        self.sign_digest(S::Digest::digest(msg))
     }
 }
 
@@ -35,6 +35,6 @@ where
     T: digest::Verifier<S::Digest, S>,
 {
     fn verify(&self, msg: &[u8], signature: &S) -> Result<(), Error> {
-        self.verify_digest(S::Digest::new().chain(msg), signature)
+        self.verify_digest(S::Digest::digest(msg), signature)
     }
 }
--- a/signature-crate/src/digest/digestable.rs
+++ b/signature-crate/src/digest/digestable.rs
@@ -1,4 +1,4 @@
-use super::Digest;
+use crate::{digest::Digest, signature::Signature};
 
 /// Marker trait for `Signature` types computable as `S(H(m))` where:
 ///
@@ -8,6 +8,7 @@ use super::Digest;
 ///
 /// For signature types that implement this trait, a blanket impl of
 /// `Signer` will be provided for all types that `impl digest::Signer`.
-pub trait Signature: crate::signature::Signature {
+pub trait Digestable: Signature {
+    /// Preferred `Digest` algorithm to use when computing this signature type.
     type Digest: Digest;
 }
--- a/signature-crate/src/digest/digestable.rs
+++ b/signature-crate/src/digest/digestable.rs
@@ -1,4 +1,10 @@
-use crate::{digest::Digest, signature::Signature};
+use crate::{
+    digest::{self, Digest},
+    error::Error,
+    signature::Signature,
+    signer::Signer,
+    verifier::Verifier,
+};
 
 /// Marker trait for `Signature` types computable as `S(H(m))` where:
 ///
@@ -11,4 +17,24 @@ use crate::{digest::Digest, signature::Signature};
 pub trait Digestable: Signature {
     /// Preferred `Digest` algorithm to use when computing this signature type.
     type Digest: Digest;
+}
+
+impl<S, T> Signer<S> for T
+where
+    S: Digestable + Signature,
+    T: digest::Signer<S::Digest, S>,
+{
+    fn sign(&self, msg: &[u8]) -> Result<S, Error> {
+        self.sign_digest(S::Digest::new().chain(msg))
+    }
+}
+
+impl<S, T> Verifier<S> for T
+where
+    S: Digestable + Signature,
+    T: digest::Verifier<S::Digest, S>,
+{
+    fn verify(&self, msg: &[u8], signature: &S) -> Result<(), Error> {
+        self.verify_digest(S::Digest::new().chain(msg), signature)
+    }
 }
--- a/signature-crate/src/digest/signer.rs
+++ b/signature-crate/src/digest/signer.rs
@@ -13,5 +13,15 @@ where
     S: Signature,
 {
     /// Sign the given prehashed message `Digest`, returning a signature.
-    fn sign(&self, digest: D) -> Result<S, Error>;
+    fn sign_digest(&self, digest: D) -> Result<S, Error>;
+}
+
+impl<S, T> crate::Signer<S> for T
+where
+    S: crate::digest::Signature,
+    T: Signer<S::Digest, S>,
+{
+    fn sign(&self, msg: &[u8]) -> Result<S, Error> {
+        self.sign_digest(S::Digest::new().chain(msg))
+    }
 }
--- a/signature-crate/src/digest/signer.rs
+++ b/signature-crate/src/digest/signer.rs
@@ -4,7 +4,7 @@
 //! For use signature algorithms that support an Initialize-Update-Finalize
 //! (IUF) API, such as ECDSA or Ed25519ph.
 
-use super::{Digest, Digestable};
+use super::Digest;
 use crate::{error::Error, signature::Signature};
 
 /// Sign the given prehashed message `Digest` using `Self`.
@@ -15,14 +15,4 @@ where
 {
     /// Sign the given prehashed message `Digest`, returning a signature.
     fn sign_digest(&self, digest: D) -> Result<S, Error>;
-}
-
-impl<S, T> crate::signer::Signer<S> for T
-where
-    S: Digestable + Signature,
-    T: Signer<S::Digest, S>,
-{
-    fn sign(&self, msg: &[u8]) -> Result<S, Error> {
-        self.sign_digest(S::Digest::new().chain(msg))
-    }
 }
--- a/signature-crate/src/digest/signer.rs
+++ b/signature-crate/src/digest/signer.rs
@@ -6,6 +6,7 @@
 
 use super::Digest;
 use crate::{error::Error, signature::Signature};
+use digest::generic_array::GenericArray;
 
 /// Sign the given prehashed message `Digest` using `Self`.
 pub trait Signer<D, S>
@@ -14,5 +15,5 @@ where
     S: Signature,
 {
     /// Sign the given prehashed message `Digest`, returning a signature.
-    fn sign_digest(&self, digest: D) -> Result<S, Error>;
+    fn sign_digest(&self, digest: GenericArray<u8, D::OutputSize>) -> Result<S, Error>;
 }
--- a/signature-crate/src/digest/signer.rs
+++ b/signature-crate/src/digest/signer.rs
@@ -4,7 +4,8 @@
 //! For use signature algorithms that support an Initialize-Update-Finalize
 //! (IUF) API, such as ECDSA or Ed25519ph.
 
-use crate::{digest::Digest, error::Error, Signature};
+use super::{Digest, Digestable};
+use crate::{error::Error, signature::Signature};
 
 /// Sign the given prehashed message `Digest` using `Self`.
 pub trait Signer<D, S>
@@ -16,9 +17,9 @@ where
     fn sign_digest(&self, digest: D) -> Result<S, Error>;
 }
 
-impl<S, T> crate::Signer<S> for T
+impl<S, T> crate::signer::Signer<S> for T
 where
-    S: crate::digest::Signature,
+    S: Digestable + Signature,
     T: Signer<S::Digest, S>,
 {
     fn sign(&self, msg: &[u8]) -> Result<S, Error> {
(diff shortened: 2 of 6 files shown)


tarcieri:
@newpavlov third time's the charm? 😉 I think this is a lot cleaner than #9 or #10, although I imagine there might need to be a little bikeshedding around the names and module structure.

tarcieri:
I was thinking `DigestableSignature` or even just `signature::Digestable` might be a good name for the marker trait too. I can try pushing a commit which uses that.

signature-crate/src/digest/signer.rs
```diff
@@ -13,5 +14,15 @@ where
     S: Signature,
 {
     /// Sign the given prehashed message `Digest`, returning a signature.
-    fn sign(&self, digest: D) -> Result<S, Error>;
+    fn sign_digest(&self, digest: D) -> Result<S, Error>;
+}
+
+impl<S, T> crate::signer::Signer<S> for T
+where
+    S: Digestable + Signature,
```
tony-iqlusion: Though this is redundant, I think it makes the bounds clearer.

tarcieri: Though this is redundant, I think it makes the bounds clearer

signature-crate/src/digest/digestable.rs
```diff
@@ -0,0 +1,14 @@
...
+/// - `S`: signature algorithm
+/// - `H`: hash (a.k.a. digest) function
+/// - `m`: message
+///
+/// For signature types that implement this trait, a blanket impl of
+/// `Signer` will be provided for all types that `impl digest::Signer`.
+pub trait Digestable: Signature {
+    /// Preferred `Digest` algorithm to use when computing this signature type.
+    type Digest: Digest;
+}
```
tarcieri: Random thought: this pattern seems generally useful... perhaps it belongs in the `digest` crate?

tarcieri:
Okay, renamed to `Digestable`, and I think this is the cleanest version yet by far.

signature-crate/src/digest/digestable.rs
newpavlov: Hm, maybe, but I think we first should see one-two similar use-cases first, before including it into `digest`.

BTW is it me or `Digestable` feels reverse in a sense? It's like "we can digest this signature", while it should be "this signature is calculated from message digest". How about `DigestSignature`? It still feels a bit off, but I don't have better ideas right now.

newpavlov:
Overall looks much better! But I am not sure if having two traits with the same name is a good idea.

Also I am not sure about `sign_digest` signature, why do we have to pass `Digest` struct and not digest output? I think something like this will be a bit more universal:
```rust
fn sign_digest(&self, digest: GenericArray<u8, D::OutputSize>) -> Result<S, Error>;
```
I think I understand that you trying to incorporate a bit more type-safety into signature, but I think that the provided variant conveys intent a bit better, i.e. that `Signer` will use message digest directly and will not mix any additional postfixes. Also shouldn't signature here bounded by `Digestable<Digest=D>` trait?

signature-crate/src/digest/digestable.rs
tarcieri: See 0ed928c which used `signature::digest::Signature` 😉

tarcieri:
> But I am not sure if having two traits with the same name is a good idea. How about `DigestSigner`?

That's actually the name Signatory used originally (there and back again!), however this approach is following [RFC356](https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md) style, which is:

In the case of `signature::digest::DigestSigner`, RFC356 suggests the name should be `signature::digest::Signer`.

I guess the spirit of RFC356 is stop worrying and love the module system 😉. The idea from an end user perspective, in the case they want to use both, is something like this:

```rust
use signature::{digest, Signer};

struct Foo<D: digest::Signer, S: Signer> { d: D, s: S }
```

...or if they want to have both traits in scope, they can alias to `DigestSigner` themselves:

```rust
use signature::{digest::Signer as DigestSigner, Signer};
```

...or:

```rust
use signature::{digest::{self, Signer as _}, Signer};
```

> Also I am not sure about `sign_digest` signature, why do we have to pass `Digest` struct and not digest output?

It'd be nice if there were a trait bound that could accommodate both, e.g. `D: Into<GenericArray<u8, D::OutputSize>>`

> Also shouldn't signature here be bounded by `Digestable<Digest=D>` trait?

No, because the underlying hash function may not be `Digestable`. It's covered in the bounds on the blanket impl.

tarcieri:
On second thought:

```rust
fn sign_digest(&self, digest: GenericArray<u8, D::OutputSize>) -> Result<S, Error>;
```

...can't work, because then `D` will be unconstrained.

newpavlov:
I think it will be fine to keep public API "flat", i.e. without any user-facing modules. Crate only contains 6 traits, so modules may be excessive. All your import options may look like a needless busy-work to users.

BTW you also forget to use `pub` on `mod digest`. Also maybe it's better to re-export `digest` crate via `pub extern crate` and not `Digest` trait directly? With the current approach by looking at docs one may decide that `Digest` is defined in the `signature` crate.

>No, because the underlying hash function may not be Digestable. It's covered in the bounds on the blanket impl.

Hm, can you provide an example? I thought that `digest::Signer` trait was for algorithms which can be describes as `S(H(m))`.

>...can't work, because then D will be unconstrained.

No, it does work without any issues, try to edit the code. Though personally I would prefer if hash function was an associated type as was described in #10 with implementations being generic over `Digest`, so users will be able to select a type describing the exact algorithm (including hash function) and not a generic one. So a hypothetical `Ed25519` will always use `Sha256` , and `Ed25519<Blake2s>` will use BLAKE-2s. Instead with the current API you will have to do a small dance around the `digest::Signer` trait.

But If you feel to, you can merge this PR as-is and we continue this discussion in new issues.

tarcieri:
> I think it will be fine to keep public API "flat", i.e. without any user-facing modules.

Yeah, `signer` and `verifier` being `pub` is left over from when I had the `SignDigest`/`VerifyDigest` nee `DigestSigner`/`DigestVerifier` traits under there. I can remove those.

> Also maybe it's better to re-export digest crate via `pub extern crate` and not `Digest` trait directly?

That's what I had here before:

https://github.com/RustCrypto/signatures/pull/11/files#diff-9c0ca6e50f797c40455a5894e3161609L18

I made the switch to permit putting `digest::Signer` and `digest::Verifier` under that module namespace as well. I really liked this change as it put all of the `Digest`-related functionality under the `digest` module which certainly made the `cfg(feature...)` gating a lot simpler.

We could flatten out the API and use `DigestSigner` and `DigestVerifier` (again, there and back again 😉 ) while doing a `pub extern crate digest`. I'm not sure how that jives with RFC356, but I imagine e.g. `signature::DigestSigner` is ok. Personally I'm a fan of the current arrangement, having tried the previous form a couple times now.

> No, it does work without any issues, try to edit the code.

Well I'll be, you're right. Okay, I can do that.

>  Though personally I would prefer if hash function was an associated type

...well, it is an associated, it's just on the signature at the algorithm level, rather than forcing every impl to redundantly re-declare it (took me awhile to get there, but that definitely seems like the right way to go to me).

> ...users will be able to select a type describing the exact algorithm (including hash function) and not a generic one

This is pretty much what you get out of the blanket impl of the `Sign` trait, aside from non-`Digestable` signatures (ok I'll change that name, it's silly 😉 ), except for things like Ed25519ph. But even then it stays out of your way unless you tried to use an invalid digest with Ed25519ph (e.g. `Sha256`, supposed to be `Sha512` 😜), in which case it would be rejected by the type system due to lack of a proper impl.

I'll note that `ed25519-dalek` now [deliberately prevents usage with BLAKE2b](https://github.com/dalek-cryptography/ed25519-dalek/issues/69#issuecomment-472174878), although with your revised `D::OutputSize`-based approach, it would be possible for it to either `impl` it for a specific hash, or for any hash whose `D::OutputSize` is 512-bits, implementer's choice.

I'll go ahead and make the `D::OutputSize` change you suggested, merge this, and submit one more followup PR to reorganize the code and change some of the names, and we can see how that works out.

newpavlov:
>it's just on the signature at the algorithm level, rather than forcing every impl to redundantly re-declare it

I mean IIUC with the current approach users will have to write code like this:
```rust
fn sign_and_send(msg: &[u8], signer: &MyAlg) -> Result<()> {
    let sig: Sig = <signer as DigestSigner<Sha256, _>>::sign(msg);
    send(sig)
}
```

While with the associated type it will be a simpler code:
```rust
// in another crate: `struct MyAlg<D: Digest = Sha256> { .. }`
fn sign_and_send(msg: &[u8], signer: &MyAlg) -> Result<()> {
    let sig: Sig = signer.sign(msg);
    send(sig)
}
// or
fn sign_and_send(msg: &[u8], signer: &MyAlg<Blake2s>) -> Result<()> {
    let sig: Sig = signer.sign(msg);
    send(sig)
}
```
Yes, algorithm crates will have a bit of boilerplate, but I think we should prioritize user ergonomics in this case.

>I'll note that ed25519-dalek now deliberately prevents usage with BLAKE2b

I think they remove genericity over `Digest` in attempt to push for 1.0, otherwise they would've to release v2.0 on `digest` v0.8 release.

tarcieri:
> I mean IIUC with the current approach users will have to write code like this:

I'm already using this approach in Signatory. The type system does not have issues resolving which impl to use when it's unambiguous. The bounds here are slightly more complex, but I don't foresee that being a problem.

Some examples:

- https://github.com/tendermint/signatory/blob/master/signatory-secp256k1/src/lib.rs#L143
- https://github.com/tendermint/yubihsm-rs/blob/master/tests/ecdsa/mod.rs#L73

newpavlov:
But there you define SHA specific extension traits which give you `sign_sha256` method, but with our current `signature` API you'll have to write code like in the snippet earlier, which looks quite unwieldy.

tarcieri:
The blanket impls are defined in a similar way, so that's a bit of a red herring. Per earlier discussions we decided it was okay to have a blessed/"happy path" digest function for each signature algorithm, which is an associated type of `DigestSignature`, and then use that for the blanket impl on the `Signer` trait.

So really the difference from Signatory will merely be it's `sign` instead of `sign_sha256` (at least for the cases of P-256 and secp256k1).

tarcieri:
Also I do plan on retrofitting this crate into Signatory, at which point I can see if all of these assumptions hold out.

newpavlov:
>to have a blessed/"happy path" digest function for each signature algorithm, which is an associated type of DigestSignature

Wait, in this case `DigestSigner` will not be able to implement `Signer` for alternative hash functions due to coherence rules. It does not look like a good trade to save a relatively tiny amount of boilerplate in algorithm implementation crates.

tarcieri:
Since there's a blanket impl, anything that implements `DigestSigner` can't impl `Signer` for any algorithms whose `Signature` type impls `DigestSigner`.

...but that's true of any blanket impl, regardless of the strategy. What the generic approach *does* allow is for that blanket impl to coexist with multiple impls of `DigestSigner`, such that, say an ECDSA P-256 signer can impl `DigestSigner` for multiple hash algorithms, e.g. `Sha256` and `Sha384`.

So it's *more* flexible/expressive, and nothing is lost.

The only thing I'm worried about is if the type resolution will still work if `DigestSigner` accepts a `GenericArray` instead of a digest as an argument. I can try retrofitting the current crate into Signatory and we can see.

newpavlov:
Ah, since signature is already tied to a hash function, it's already generic over `Digest`, so my snippet from earlier should look like this:
```rust
// `Sig<D: Digest = Sha256>` 
fn sign_and_send(msg: &[u8], signer: &MyAlg) -> Result<()> {
    let sig: Sig = signer.sign(signer, msg);
    send(sig)
}
// or
fn sign_and_send(msg: &[u8], signer: &MyAlg) -> Result<()> {
    let sig: Sig<Blake2s> = signer.sign(signer, msg);
    send(sig)
}
```
Correct?

tarcieri:
In the current implementation, `Signature` is a trait and the `Digest` to use is an associated type.

When defining the signature types for the relevant algorithms, we could potentially make them generic around the digest and use the generic parameter as the associated type. So yes, what you're doing above should be possible.

newpavlov:
>The only thing I'm worried about is if the type resolution will still work if DigestSigner accepts a GenericArray instead of a digest as an argument.

Without `Digestable<Digest=D>` bound on `S` I don't think it will. What are counter-examples to such bound?

tarcieri:
I encounter a couple problems with this approach trying to retrofit it into Signatory. Neither are showstoppers, but both are annoying.

The first:

```
   = note: conflicting implementation in crate `signature`:
           - impl<S, T> signatory::Signer<S> for T
             where S: signatory::DigestSignature, T: signatory::DigestSigner<<S as signatory::DigestSignature>::Digest, S>;
   = note: upstream crates may add new impl of trait `signatory::DigestSignature` for type `signatory::ed25519::Signature` in future versions
```

Seems it won't be possible for types to implement both `Signer` and `DigestSigner` due to the blanket `impl`. They will need to use separate types per algorithm. Unless there's some way around this particular error I'm unaware of...

Additionally, Ed25519Ph is domain separated and takes a `context: Option<&[u8]>` parameter. So I guess a separate question is if we want to provide traits for these sorts of algorithms, and if so, what do they look like.

I can make a separate issue for the latter.

@newpavlov re: your question, still slogging through a Signatory retrofit. I'll see if anything comes up.

PR #13, opened 2019-04-01
Factor Digest-related traits out of 'digest' module

tarcieri:
...allowing us to (once again) `pub extern crate digest`.

This PR undoes many of the changes from #11 and once again flattens out the crate's API, adding a `Digest*` prefix to all traits that deal with `Digest` types, and re-exporting them from the crate root.

It also renames `Digestable` to `DigestSignature`, which is more clear as to its purpose.


tarcieri:
@newpavlov I think this PR contains changes you suggested. Will go ahead and merge, but feel free to leave some follow-up comments and I can do another round of changes before a release.

Would like to start spiking out the `ecdsa` and `ed25519` crates next.

newpavlov:
Looks good!

You probably also can improve documentation a bit, e.g. you can move HSM mention from `Signer` to second line and change "computable as S(H(m))" to "computable as a signature of message digest" and move `S(H(m))` to the detailed documentation, but it can be done later without any problems.

tarcieri:
@newpavlov sounds good

PR #14, opened 2019-04-01
Add `try_*` methods to `*Signer` traits

tarcieri:
For many signature providers (i.e. most software implementations), it's not possible for an error to occur when creating a signature. Support for handling errors which occur when computing signatures is intended for use with cloud KMS, HSMs, or other hardware tokens where things like I/O errors are possible.

To simplify usage in the event that only software signers are in use, this commit splits the `sign` API into `Result`-based `try_sign` method and another `sign` method that always assumes success and unwraps the result of the former method.


newpavlov:
Looks good! Though, I don't know if we should use a simple `unwrap` here or `expect`. I've seen both approaches used in the wild.

tarcieri:
`expect` is a good call, I can switch to that

PR #16, opened 2019-06-05
Replace Digest-related blanket impls with methods

tarcieri:
The current "clever" blanket impl (for which I'm entirely responsible!) seems to have bounds which prevent anything from implementing `Signer` or `Verifier` when the `digest` feature is enabled:

```
   = note: conflicting implementation in crate `signature`:
           - impl<S, T> signature::signer::Signer<S> for T
             where S: signature::signature::DigestSignature, T: signature::signer::DigestSigner<<S as signature::signature::DigestSignature>::Digest, S>;
   = note: upstream crates may add new impl of trait `signature::signature::DigestSignature` for type `signatory::ed25519::signature::Signature` in future versions
   = note: downstream crates may implement trait `signature::signer::DigestSigner<_, signatory::ed25519::signature::Signature>` for type `ed25519::Ed25519Signer`
```

(I don't even understand the `note: downstream crates...` issue as I wasn't aware it was possible for a downstream crate to impl a trait in one upstream dependency for a type in another upstream dependency, however the `note: upstream crates may add new impl of trait DigestSigner` seems like a dealbreaker)

This commit, while a bit ugly, at least eliminates all of the complexity around bounds on the blanket impl by entirely removing the blanket impl, and adding an additional method to precompute the digest.

I'm not entirely happy with this and think the method names are a bit confusing and not descriptive enough, but it does have the following rather nice properties:

- Easy to reason about: no blanket impls
- Allows the same type to impl both the `Digest` and non-`Digest` forms
  of `Signer` and `Verifier` simultaneously.

There are a lot of potential directions this could go in order to improve the API, potentially splitting it into two traits rather than one (which gets us back to the `Sha*Signer`/`Sha*Verifier` traits which
Signatory used originally), however I think this is the MVP for actually using the `signature` crate in Signatory, so I'd like to start with this and then iterate towards a better API.

### DIFF
--- a/signature-crate/src/signature.rs
+++ b/signature-crate/src/signature.rs
@@ -22,19 +22,3 @@ pub trait Signature: AsRef<[u8]> + Debug + Sized {
         self.as_slice().into()
     }
 }
-
-/// Marker trait for `Signature` types computable as `S(H(m))`
-///
-/// - `S`: signature algorithm
-/// - `H`: hash (a.k.a. digest) function
-/// - `m`: message
-///
-/// For signature types that implement this trait, a blanket impl of
-/// `Signer` will be provided for all types that `impl DigestSigner`
-/// along with a corresponding impl of `Verifier` for all types that
-/// `impl DigestVerifier`.
-#[cfg(feature = "digest")]
-pub trait DigestSignature: Signature {
-    /// Preferred `Digest` algorithm to use when computing this signature type.
-    type Digest: digest::Digest;
-}
--- a/signature-crate/src/signer.rs
+++ b/signature-crate/src/signer.rs
@@ -1,10 +1,7 @@
 //! Traits for generating digital signatures
 
 #[cfg(feature = "digest")]
-use crate::{
-    digest::{generic_array::GenericArray, Digest},
-    signature::DigestSignature,
-};
+use crate::digest::{generic_array::GenericArray, Digest};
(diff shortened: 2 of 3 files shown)


tarcieri:
I am going to attempt to finish retrofitting `signature` into Signatory using this change, and if it works out, I'd like to merge this and cut another release, and then we can discuss paths forward which provide a nicer API. Perhaps we should revisit some of the approaches @newpavlov was originally proposing for layering an object-safe API on top of a concrete API.

tarcieri:
In other news, it seems our `DigestSigner` trait is incompatible with `ed25519-dalek`'s `sign_prehashed`, since it takes a `Digest` instance as input:

https://docs.rs/ed25519-dalek/1.0.0-pre.1/ed25519_dalek/struct.Keypair.html#method.sign_prehashed

tarcieri:
It looks like this (and one additional PR) will be sufficient to get Signatory to compile against the `signature` crate.

Here is my tentative plan:

- Merge this, PR the additional necessary changes (see `DigestSigner` comment above), merge that, and release `v0.2.0-pre.1`
- Revisit the blanket impl. I think it will work if constrained to particular digests, which in practice for common signature algorithms is generally the case.

This API is bad and shouldn't go out in a final release, but Signatory is a great testbed for how this stuff actually works in practice, and I want to get it working with some released crate until we can discuss options for the blanket impl.

PR #18, opened 2019-06-06
signature_derive: Custom derive support for Signer/Verifier

tarcieri:
We've explored several different possible trait bounds for a blanket impl of `Signer` for `DigestSigner` (and likewise for `Verifier`).

Unfortunately, the approach we netted out at in #12 does not permit anything but the blanket impl, and had to be replaced (with something I think is suboptimal and messy) in #16.

It's definitely still worth exploring if there's a way to define the bounds of the default impl that works, however there is one approach we haven't explored yet: a procedural macro.

This approach feels like a bit of a hack, but gets us to where we originally wanted to be with the blanket impl API-wise. The impl it derives uses the same approach and bounds as the blanket impl did: the `Digest` to use is sourced from an associated type of the `DigestSignature` trait, which means the digest to use is both automatic and ensures the type deriving the trait supports the expected `Digest` for the given signature.

The implementation uses `synstructure` which simplifies both handling of generics and testing the output of the proc macro.

Additionally, it includes a complete integration test of the derived code which ensures it works as expected.

It's gated under a cargo feature and disabled-by-default. This should make it unobtrusive for downstream crates which don't need its functionality.


tarcieri:
I will try integrating this change into `signatory` and see how it pans out when used with real-world `Signer`/`Verifier` impls.

tarcieri:
I've managed to update Signatory using these changes, and I really like the result.

I'd like to ship this in `signature` crate `v0.2.0`, and then ship a release of `signatory` which is based on this crate.

We can evaluate whether `signature_derive` can be eliminated by changing the API to better leverage the type system in `signature` v0.3 (or later).

PR #23, opened 2019-08-09
ed25519: Initial implementation (closes #22)

### LINKED ISSUE #22: ed25519 crate
Now that the `signature` crate is released in some form and being used, the next step I think makes sense is to work on the `ed25519` crate (which @newpavlov has already reserved).

I think the best way to keep it simple for an initial release would be to focus exclusively on an `ed25519::Signature` type which impls the [`signature::Signature`](https://docs.rs/signature/latest/signature/trait.Signature.html) trait and can be shared across crates that produce and consume Ed25519 signatures in conjunction with the [`signature::Signer`](https://docs.rs/signature/latest/signature/trait.Signer.html) and [`signature::Verifier`](https://docs.rs/signature/latest/signature/trait.Signer.html) traits.

A unified public key type would be nice as well, although the `signature` crate presently has no traits relating to public keys, so this seems like a good issue to punt on until the question of the relationship between the `signature` crate and public keys is addressed.

An `ed25519::Seed` type containing the derivation (via SHA-512) secret for the full keypair might be interesting as well, but also something that seems worthwhile to punt on in the short term.

tarcieri:
Basic implementation with an initial `ed25519::Signature` type which impls the `signature::Signature` trait.

### DIFF
--- a/ed25519/Cargo.toml
+++ b/ed25519/Cargo.toml
@@ -1,13 +1,18 @@
 [package]
-name = "ed25519"
-version = "0.0.0"
-authors = ["RustCrypto Developers"]
-license = "Apache-2.0 OR MIT"
-description = "Edwards Digital Signature Algorithm (EdDSA) over Curve25519 (as specified in RFC 8032)"
+name          = "ed25519"
+version       = "0.0.0"
+authors       = ["RustCrypto Developers"]
+license       = "Apache-2.0 OR MIT"
+description   = "Edwards Digital Signature Algorithm (EdDSA) over Curve25519 (as specified in RFC 8032)"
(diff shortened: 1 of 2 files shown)


tarcieri:
> Regarding `Cargo.toml` formatting: isn't it quite uncommon to align values in such fashion?

Well, I did the same thing in the `signature` crate 😉 

https://github.com/RustCrypto/signatures/blob/master/signature-crate/Cargo.toml

Could go either way on it (I like things lined up, but it seems less popular to do in Rust than other languages), but if nothing else I guess I'd prefer they be consistent.

tarcieri:
Note: CI was breaking on `serde` support, which I could also go either way on but punted on for now.

Issue #25, opened 2019-08-19
RSA signature concerns


roblabla:
In RSA, there are multiple schemes to sign/verify, mainly PKCS1 v1.5 and PSS (although others exists). I was looking at how to integrate `signature` into the RSA crate, and was a bit confused at the lack of a Padding argument or type in `Signer`/`Verifier`. Is the idea to implement `Verifier` on a type that is already aware of the padding scheme?

newpavlov:
In my understanding `rsa` types should be generic over a hypothetical `RsaPadding` trait, e.g. `RsaPrivateKey<P: RsaPadding>`, which in turn will:
```rust
impl<P: RsaPadding> Signer for RsaPrivateKey<P> { .. }
```

tarcieri:
@roblabla a hypothetical `rsa` crate should probably follow the basic module structure I have in the `yubihsm-rs` crate (note that it incorporates encryption too, so it's a bit more complex than is needed for just the signature use case, although I think an `rsa` crate needs to be as well):

https://github.com/tendermint/yubihsm-rs/tree/develop/src/rsa

As @newpavlov mentioned, the `signature` crate is agnostic to the underlying implementation details of each signature algorithm, e.g. a hypothetical ECDSA crate needs to be generic over elliptic curve choice (P-256, P-384, secp256k1) and these details are irrelevant to RSA.

For an `rsa` crate I would suggest making a distinct signature type for RSASSA PKCS#1v1.5 and PSS, which are potentially unified by a common `rsa::Signature` trait which is in turn bounded by `signature::Signature`.

Note that PSS likely needs to be generic around a `Digest` suitable as a Mask Generating Function (MGF).

tarcieri:
Note: renamed the title to something a bit more generic as I think these are important concerns to capture somewhere

dignifiedquire:
I think I agree with the suggestions from @tarcieri on this, @roblabla I am very happy to break the rsa crates interface in order to get a better interface, so if you want to write up a proposal based on this for how the api could look like that would be great.

roblabla:
I picked this up again, trying to fit `signature` into the RSA crate. The more I look into it the more confused I am by the traits. Here are a couple of things I noticed:

[...]

PR #32, opened 2019-10-10
signature crate: 1.0 stabilization proposal

tarcieri:
The motivation for 1.0 stabilization of the `signature` crate is the upcoming 1.0 release of `ed25519-dalek`. In order to promote Ed25519 interoperability, it would be great if `ed25519-dalek` could use the traits from this crate along with the `ed25519::Signature` type from the `ed25519` crate. See the relevant issue:

https://github.com/dalek-cryptography/ed25519-dalek/issues/80

To get there, I think we need to do a 1.0 release of this crate, as well as the `ed25519` crate.

The main impediment towards doing so is the `digest` crate is presently stuck at v0.8. It would be nice to be able to continue upgrading it, especially to a 1.0 release, but that would otherwise be a SemVer
breaking change.

To allow agility around `digest`, and `signatory_derive` which depends on it, this commit places access to both under the `digest-preview` and `derive-preview` Cargo features respectively, and calls them out as not covered under SemVer and subject to change, but breaking changes will be done with a minor version bump.


signature-crate/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name          = "signature"
 description   = "Traits for cryptographic signature algorithms (e.g. ECDSA, Ed25519)"
-version       = "0.3.0" # Also update html_root_url in lib.rs when bumping this
+version       = "1.0.0-pre" # Also update html_root_url in lib.rs when bumping this
```
tarcieri: Note: if we do this, I'd like to do a series of prereleases (e.g. `1.0.0-pre.0`, `1.0.0-pre.1`) to work through issues encountered with `ed25519-dalek`, `signatory`, and potentially `rsa` as well /cc @roblabla

tarcieri:
One thing I'd definitely like to do before a 1.0 release is improve the state of the documentation, both for consumers of the trait, and for crates that impl the provided traits. Per #25 there are a number of undocumented design decisions which need to be adequately described.

tarcieri:
Going to move forward with an initial 1.0 prerelease using this plan. If anyone would like to get any changes in prior to a final 1.0, I don't plan on being too hasty in cutting that.

tarcieri:
Note: final comment period for shipping `signature` 1.0 in ~2 weeks here: https://github.com/RustCrypto/traits/issues/78

PR #57, opened 2019-12-05
elliptic-curve-crate: Extract from `ecdsa` crate

tarcieri:
The `ecdsa` crate contains types, traits, and a complete registry for common Weierstrass curves which is reusable for both the signing and encryption use case.

This extracts the reusable functionality into the `elliptic-curve` crate, with the goal of rewriting the duplicated parts of the current `ecdsa` crate to use types from the `elliptic-curve` crate, and once
that's done, finding a better home, e.g. https://github.com/RustCrypto/traits

It seems we collectively own the following crates:

- `curve25519`
- `ed25519`
- `p256`
- `p384`
- `p521`
- `k256`

...so it would be nice to extract the curve registry in these crates into those crates respectively, e.g. under a hypothetical https://github.com/RustCrypto/elliptic-curves

But for now, this structures the crate the same way the `ecdsa` crate was previously structured: with an all-in-one curve registry contained within the crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 members = [
     "ecdsa",
     "ed25519",
+    "elliptic-curve-crate",
     "signature-crate",
     "signature-crate/signature_derive"
 ]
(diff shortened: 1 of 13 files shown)


tarcieri:
Will take a look at adapting the `ecdsa` crate to use this.

Please feel free to leave comments / reviews in the meantime.

PR #63, opened 2019-12-07
secp256r1 point compression and decompression

str4d:
Implements the necessary field arithmetic to support these operations.


elliptic-curve-crate/src/p256.rs
```diff
@@ -0,0 +1,210 @@
...
+//! # Status
+//!
+//! Currently, no actual group operations are implemented. Only point compression and
+//! decompression is supported.
+
+mod field;
+mod util;
+
+#[cfg(test)]
+mod test_vectors;
```
tarcieri: My personal inclination on test vectors is they should go in `tests/` unless you'd like to expose them under a cargo feature, e.g. `test-vectors`, so implementations can use them as part of an initial self-test (ala FIPS).

Perhaps add a `test-vectors` cargo feature and make this `#[cfg(any(feature = "test-vectors", test))]`?

elliptic-curve-crate/Cargo.toml
```diff
@@ -23,13 +23,21 @@ optional = true
 version = "0.12"
 default-features = false
 
+[dependencies.subtle]
+version = "2.2.2"
```
tarcieri: Can you make this:

```suggestion
version = "2.2.2"
default-features = false
```

elliptic-curve-crate/src/p256.rs
str4d: I left them inside the module as I figured they would be moved out with the module itself into the `p256` crate. I can do either of the above if you want (these test vectors were what I used to identify a carry bug during development).

tarcieri: I think it'd be good to optionally expose them to downstream consumers, and perhaps eventually provide a self-test

PR #64, opened 2019-12-07
p256: Improve encoding APIs

str4d:
The compressed and uncompressed encoding APIs now return the corresponding precise type, which can be converted into a `PublicKey` via `PublicKey::from`.


tarcieri:
Looks good, and like there are probably a few more of these sort of `From` impls which would be nice, but I'll save that for another PR.

PR #90, opened 2020-06-28
ecdsa::curve::secp256k1::RecoverableSignature

tarcieri:
Adds a signature type for Ethereum-style recoverable signatures.

### DIFF
--- a/ecdsa/src/curve/secp256k1/recoverable_signature.rs
+++ b/ecdsa/src/curve/secp256k1/recoverable_signature.rs
@@ -1,0 +1,105 @@
+//! Ethereum-style "recoverable signatures"
+
+use super::FixedSignature;
+use crate::{
+    generic_array::{typenum::U64, GenericArray},
+    Error,
+};
+use core::{
+    convert::{TryFrom, TryInto},
+    fmt::{self, Debug},
+};
+
+/// Size of an Ethereum-style recoverable signature in bytes
+pub const SIZE: usize = 65;
+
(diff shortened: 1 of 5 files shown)


ecdsa/src/curve/secp256k1/recoverable_signature.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+
+/// Size of an Ethereum-style recoverable signature in bytes
+pub const SIZE: usize = 65;
+
+/// Ethereum-style "recoverable signatures" which allow for the recovery of
+/// the signer's [`PublicKey`] from the signature itself.
+///
+/// These consist of [`FixedSignature`] followed by a 1-byte [`RecoveryId`]
+/// (65-bytes total)
```
tarcieri, line 20: It'd be nice to reference some kind of specification here. Unfortunately I wasn't able to find one. Anybody know of one?

tarcieri: The only signature specification I was able to find is [EIP-2098](https://eips.ethereum.org/EIPS/eip-2098) which describes a new 64-byte signature format which is able to take advantage of low-s normalization to pack the recovery ID into the `s` value.

It seems there isn't a definitive spec on this format anywhere otherwise. However I've double checked the message structure against a few places where it's documented, like this (see "Extracting the Signature Parameters" section):

https://programtheblockchain.com/posts/2018/02/17/signing-and-verifying-messages-in-ethereum/

PR #119, opened 2020-08-06
Make `normalize_low()` pure

fjarri:


tarcieri:
Looks ok. The `codecov` failure is a misconfiguration (old config that needs access to a token) and a merge will succeed.

PR #133, opened 2020-08-26
ecdsa: implement RFC6979 ephemeral scalar generation

### LINKED ISSUE #124: Deterministic ECDSA support (RFC 6979 and more)
Currently the `ecdsa` crate's `ecdsa::Signer` type only impls `RandomizedSigner` and `RandomizedDigestSigner`, and uses a `k` value derived verbatim from the RNG. This approach is historically fraught with problems and there are better solutions.

# RFC 6979

Due to an ongoing litany of ECDSA secret key recovery vulnerabilities owing to accidental repetitions of the ECDSA `k` (nonce) value for signing multiple messages under the same secret key, deterministic methods of generating `k` like [RFC 6979](https://tools.ietf.org/html/rfc6979) have grown popular. Additionally, a deterministic method would allow us to sign without an RNG, i.e. impl the `Signer` and `DigestSigner` traits.

# draft-mattsson-cfrg-det-sigs-with-noise

Unfortunately, deterministic `k` generation makes it easy for an attacker with physical access who can inject faults to potentially induce repeated signatures under the same `k` value. An attacker who can inject a fault in the calculation of the ECDSA `z` scalar from the message digest some time after `k` has been computed (or otherwise inject faults in the subsequent arithmetic on `z`) can leverage this to recover the private key.

[draft-mattsson-cfrg-det-sigs-with-noise](https://tools.ietf.org/html/draft-mattsson-cfrg-det-sigs-with-noise-02) provides a hybrid approach that bolsters an RFC 6979-like construction with additional randomness to ensure `k` is still uniformly random per-message in the event of an RNG failure, but also uniquely random every time in the event the RNG is working. Ideally we can change the `RandomizedSigner` and `RandomizedDigestSigner` traits to use this sort of approach.

tarcieri:
Adds a deterministic signing mode based on RFC6979. Closes #124.

~~This could be further extended to bolster the security when using `RandomizedSigner` using the method described in Section 3.6.~~

Edit: went ahead and made `RandomizedDigestSigner` use an RFC6979-style derivation but supplying some additional data from the RNG as input. It was easy enough.

### DIFF
--- a/ecdsa/src/signer/rfc6979.rs
+++ b/ecdsa/src/signer/rfc6979.rs
@@ -1,0 +1,115 @@
+//! Support for computing deterministic ECDSA ephemeral scalar (`k`) using
+//! the method described in RFC 6979 rules (Section 3.2):
+//! <https://tools.ietf.org/html/rfc6979#section-3>
+
+use elliptic_curve::{
+    digest::{BlockInput, FixedOutput, Reset, Update},
+    generic_array::GenericArray,
+    ops::Invert,
+    scalar::NonZeroScalar,
+    zeroize::{Zeroize, Zeroizing},
+    Arithmetic, ElementBytes, FromBytes, FromDigest,
+};
+use hmac::{Hmac, Mac, NewMac};
+
+/// Generate ephemeral scalar `k` from the secret scalar and a digest of the
+/// input message.
+pub(super) fn generate_k<C, D>(
+    secret_scalar: &NonZeroScalar<C>,
+    msg_digest: D,
+    additional_data: &[u8],
+) -> Zeroizing<NonZeroScalar<C>>
+where
+    C: Arithmetic,
+    C::Scalar: FromDigest<C> + Invert<Output = C::Scalar> + Zeroize,
+    D: BlockInput<BlockSize = C::ElementSize>
+        + FixedOutput<OutputSize = C::ElementSize>
+        + Clone
+        + Default
+        + Reset
+        + Update,
+    ElementBytes<C>: Zeroize,
+{
+    let x = Zeroizing::new(secret_scalar.to_bytes());
+    let h1: ElementBytes<C> = C::Scalar::from_digest(msg_digest).into();
+    let mut hmac_drbg = HmacDrbg::<D>::new(&*x, &h1, additional_data);
+
+    loop {
+        let k = NonZeroScalar::from_bytes(&hmac_drbg.next());
+
+        if k.is_some().into() {
+            return Zeroizing::new(k.unwrap());
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


ecdsa/src/signer/rfc6979.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+}
+
+/// Internal implementation of `HMAC_DRBG` as described in NIST SP800-90A:
+/// <https://csrc.nist.gov/publications/detail/sp/800-90a/rev-1/final>
+///
+/// This is a HMAC-based deterministic random bit generator used internally
+/// to compute a deterministic ECDSA ephemeral scalar `k`.
+// TODO(tarcieri): generalize and extract this into the `hmac` crate?
+struct HmacDrbg<D>
```
tarcieri, line 58: @newpavlov it might be interesting to extract or reimplement this in the `hmac` crate (possibly as a `CryptoRng`?)

newpavlov: I think it will be better to implement it in a separate crate and it would fit nicely into the currently empty [CSRNGs repository](https://github.com/RustCrypto/CSRNGs). And it looks like @sorpaas has already [published](https://crates.io/crates/hmac-drbg) such crate (although without implementing the `rand_core` traits).

@sorpaas
Would you be interested in transferring `hmac-drbg` crate to this organization?

tarcieri: Oh neat! Unfortunately it's `hmac` dependency is a version behind or otherwise it looks like what I need.

ecdsa/src/signer/rfc6979.rs
```diff
@@ -0,0 +1,123 @@
...
+        + Reset
+        + Update,
+    ElementBytes<C>: Zeroize,
+{
+    let x = Zeroizing::new(secret_scalar.to_bytes());
+
+    // TODO(tarcieri): don't panic (i.e. unwrap)! add/use trait for reducing digests mod q
+    let h1: ElementBytes<C> = C::Scalar::from_bytes(&msg_digest.finalize())
+        .unwrap()
+        .into();
```
tarcieri, line 40: TODO: Need a `Scalar::from_digest`-style trait here.

ecdsa/src/signer/rfc6979.rs
tarcieri: Opened a PR to bump the `hmac` dependency here: https://github.com/tarcieri/rust-hmac-drbg/pull/1

tarcieri:
Removed draft and WIP.

This still doesn't have tests, but due to the nature of RFC6979 it's a bit tricky to test in isolation. It would require implementing the rudiments of a particular elliptic curve and writing tests against it.

It will be easier to test in conjunction with particular elliptic curve implementations.

PR #151, opened 2020-09-03
ecdsa: add CheckSignatureBytes trait + NonZeroScalar components

tarcieri:
Adds a `CheckSignatureBytes` trait with a default impl that ensures neither the `r` nor `s` components of a signature are all-zero.

Includes a blanket impl of `CheckSignatureBytes` for all curves with the `Arithmetic` trait defined (and thus access to `C::Scalar`). This performs a better check that both `r` and `s` do not overflow the curve's order and are non-zero (using `NonZeroScalar`).

Together these allow conditionally defining `ecdsa::Signature::r` and `::s` methods which permit infallible access to `NonZeroScalar<C>`, rather than the previous implementation which returns bytes.

All of the above is a compromise which allows for crates which do not provide a curve arithmetic backend to still assert some checks on the signature values, but also ensure that signatures are well-formed, in-range scalars "automatically" when arithmetic is available.

### DIFF
--- a/ecdsa/src/lib.rs
+++ b/ecdsa/src/lib.rs
@@ -65,11 +65,11 @@ pub use sign::SigningKey;
 pub use verify::VerifyKey;
 
 use core::{
-    convert::TryFrom,
+    convert::{TryFrom, TryInto},
     fmt::{self, Debug},
     ops::Add,
 };
-use elliptic_curve::{Arithmetic, ElementBytes, FromBytes};
+use elliptic_curve::{scalar::NonZeroScalar, Arithmetic, ElementBytes, FromBytes};
 use generic_array::{sequence::Concat, typenum::Unsigned, ArrayLength, GenericArray};
 
 /// Size of a fixed sized signature for the given elliptic curve.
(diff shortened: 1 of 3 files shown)


ecdsa/src/lib.rs
```diff
@@ -214,39 +229,92 @@ where
...
+}
+
+impl<C> CheckSignatureBytes for C
+where
+    C: Curve + Arithmetic,
+    SignatureSize<C>: ArrayLength<u8>,
+{
+    /// When curve arithmetic is available, check that the scalar components
+    /// of the signature are in range.
+    fn check_signature_bytes(bytes: &SignatureBytes<C>) -> Result<(), Error> {
```
tarcieri, line 308: The reason why this checks the signature as a whole rather than the individual components is it might make sense in some cases to have special-case handling of the individual components. The main case that comes to mind is EIP-2098 signatures. But this may be of dubious value, and it may still make more sense to have a "check scalar in range" trait which can be available even if a full curve arithmetic implementation is not.

PR #243, opened 2021-02-12
Change HmacDrbg to support variable output size

rvolgers:
This is simply a matter of iterating the hashing of self.v until
sufficient output bytes have been produced.

This is necessary for the upcoming DSA implementation.

### DIFF
--- a/ecdsa/src/rfc6979.rs
+++ b/ecdsa/src/rfc6979.rs
@@ -34,7 +34,9 @@ where
     x.zeroize();
 
     loop {
-        if let Some(k) = NonZeroScalar::from_repr(hmac_drbg.next()) {
+        let mut tmp = FieldBytes::<C>::default();
+        hmac_drbg.generate_into(&mut tmp);
+        if let Some(k) = NonZeroScalar::from_repr(tmp) {
             return Zeroizing::new(k);
         }
     }
@@ -87,17 +89,18 @@ where
     }
 
     /// Get the next `HMAC_DRBG` output
-    pub fn next(&mut self) -> GenericArray<u8, D::OutputSize> {
-        self.k.update(&self.v);
-        let t = self.k.finalize_reset().into_bytes();
+    pub fn generate_into(&mut self, out: &mut [u8]) {
+        for out_chunk in out.chunks_mut(self.v.len()) {
+            self.k.update(&self.v);
+            self.v = self.k.finalize_reset().into_bytes();
+            out_chunk.copy_from_slice(&self.v[..out_chunk.len()]);
+        }
 
-        self.k.update(&t);
+        self.k.update(&self.v);
         self.k.update(&[0x00]);
         self.k = Hmac::new_varkey(&self.k.finalize_reset().into_bytes()).unwrap();
-        self.k.update(&t);
+        self.k.update(&self.v);
         self.v = self.k.finalize_reset().into_bytes();
-
-        t
     }
 }
 


ecdsa/src/rfc6979.rs
```diff
@@ -87,17 +89,18 @@ where
     }
 
     /// Get the next `HMAC_DRBG` output
-    pub fn next(&mut self) -> GenericArray<u8, D::OutputSize> {
-        self.k.update(&self.v);
-        let t = self.k.finalize_reset().into_bytes();
+    pub fn generate_to_slice(&mut self, out: &mut [u8]) {
```
tarcieri, line 92: We've typically used a `*_into` naming convention for this sort of API elsewhere in the project. How about:

```suggestion
    pub fn generate_into(&mut self, out: &mut [u8]) {
```

tarcieri: There's also `XofReader` for precedent: https://docs.rs/digest/0.9.0/digest/trait.XofReader.html

rvolgers: Let's go with `generate_into` then.

rvolgers:
@tarcieri By the way, the next step is splitting this to a separate crate. Would it be acceptable to keep it in the same repository though? It is not a signature algorithm obviously, but it's most easily tested as part of ecdsa / dsa.

tarcieri:
@rvolgers yeah that's fine. I registered the [`rfc6979` crate](https://crates.io/crates/rfc6979)

rvolgers:
RFC6979 is about DSA / ECDSA. Do you propose to move the k generation into this crate as well, then? That probably means adding separate features for DSA and ECDSA.

rvolgers:
Actually, that seems fine. It sidesteps a number of other complicating factors.

tarcieri:
I think everything in the current `ecdsa::rfc6979` module can be extracted, including the rejection sampling. However, the `ecdsa` crate should depend on it, which means elliptic curve-specific types like `NonZeroScalar` and `FieldBytes` should remain in the `ecdsa` crate.

rvolgers:
@tarcieri I'm running into some trouble with the `FromDigest` trait and stuff from `dev.rs`. I don't fully understand why this exists at all, it seems like a lot of ad-hoc code for what should be a generic "from bytes" operation. But I guess I'm missing something. Should I just move all that stuff to the new crate?

tarcieri:
> I'm running into some trouble with the FromDigest trait and stuff from dev.rs. I don't fully understand why this exists at all, it seems like a lot of ad-hoc code for what should be a generic "from bytes" operation.

Can you be more specific? Are you talking about this line?

```rust
    let h1 = Scalar::<C>::from_digest(msg_digest).to_repr();
```

If so, it corresponds directly to this pseudocode from [RFC6979 Section 2.4](https://tools.ietf.org/html/rfc6979#section-2.4):

```
       H(m) is transformed into an integer modulo q using the bits2int
       transform and an extra modular reduction:

          h = bits2int(H(m)) mod q

       As was noted in the description of bits2octets, the extra modular
       reduction is no more than a conditional subtraction.
```

The modular reduction should happen in the `ecdsa` and `dsa` crates, IMO. The `rfc6979` crate can accept a mod-reduced input, and document that as part of the contract of the function.

This needs to happen anywhere you see anything specific to DSA or ECDSA: the `rfc6979` crate can only implement the functionality which is non-specific to a particular algorithm.

rvolgers:
I'm talking about refactoring trouble, as in, I move stuff and the compiler yells at me. This is boring work and I'm trying to limit the amount of time I'm spending to figure out random implementation details I don't care about, of which there are a lot (cargo features, large amount of generics etc).

In particular, the testcase has a dependency on the `FromDigest` impl for `MockCurve`, and I can't import that as it creates a circular dependency. So either I have to move that to the new crate also, or change the test. I didn't fully understand what it's for, especially since it seems to only be implemented for `MockCurve`, unless other implementations live elsewhere, in which case, the trait should probably also live there and not in ecdsa or the new crate?

rvolgers:
I'll put up a draft PR soon, it's probably no use talking about it like this.

tarcieri:
Yeah, pushing up some actual code would help.

> In particular, the testcase has a dependency on the FromDigest impl for MockCurve, and I can't import that as it creates a circular dependency. So either I have to move that to the new crate also, or change the test. I didn't fully understand what it's for, especially since it seems to only be implemented for MockCurve, unless other implementations live elsewhere, in which case, the trait should probably also live there and not in ecdsa or the new crate?

Unlike DSA, ECDSA needs to support multiple elliptic curves, hence the generics. `MockCurve` is used for testing algorithms which are implemented generically over elliptic curves. The other impls of `FromDigest` live within the respective elliptic curve crates which provide concrete implementations. 

The `FromDigest` trait describes an ECDSA-specific algorithm as defined in ANSI X9.62-2005 and absolutely should remain in the `ecdsa` crate. It could perhaps be better documented in this regard.

In porting over the existing tests, you should operate on the raw output of `bits2int(H(m)) mod q`, as everything before that is an ECDSA-specific concern.

PR #244, opened 2021-02-13
Clarify purpose of FromDigest trait

rvolgers:
This is my best understanding from what I was told in another issue.

### DIFF
--- a/ecdsa/src/hazmat.rs
+++ b/ecdsa/src/hazmat.rs
@@ -145,8 +145,12 @@ pub trait DigestPrimitive: Curve {
 
 /// Instantiate this type from the output of a digest.
 ///
-/// This can be used for implementing hash-to-scalar (e.g. as in ECDSA) or
-/// hash-to-curve algorithms.
+/// This trait is intended for use in ECDSA and should perform a conversion
+/// which is compatible with the rules for calculating `h` from `H(M)` set out
+/// in RFC6979 section 2.4. This conversion cannot fail.
+///
+/// This trait may also be useful for other hash-to-scalar or hash-to-curve
+/// use cases.
 #[cfg(feature = "digest")]
 #[cfg_attr(docsrs, doc(cfg(feature = "digest")))]
 pub trait FromDigest<C: Curve> {


tarcieri:
The `FromDigest` trait is not specifically related to RFC 6979.

The main use case is as a core part of the ECDSA algorithm, however the operation it performs, a hash-to-scalar transformation based on a modular output of a hash function, is the same as the operation described in [RFC6979 Section 2.4](https://tools.ietf.org/html/rfc6979#section-2.4):

```
       H(m) is transformed into an integer modulo q using the bits2int
       transform and an extra modular reduction:

          h = bits2int(H(m)) mod q

       As was noted in the description of bits2octets, the extra modular
       reduction is no more than a conditional subtraction.
```

ECDSA itself performs a modular reduction of the output of a hash function in order to compute a scalar 𝑒 from `H(m)`, which is subsequently used for both signing and verification.

rvolgers:
And that's why it should remain in `ecdsa` instead of being moved to `rfc6979`, right?

It's good additional context, but I don't see anything actionable for me to change. The RFC6979 explanations of anything are much clearer than the ones in the DSA standard, and the point is to be clear about the expected semantics.

tarcieri:
Yes, that's why it should be in `ecdsa` and the `rfc6979` crate should consume the reduced outputs.

I think the wording is a bit confusing in that it cites RFC 6979, but unfortunately the primary documentation for ECDSA (e.g. ANSI X9.62) is paywalled.

ecdsa/src/hazmat.rs
```diff
@@ -145,8 +145,12 @@ pub trait DigestPrimitive: Curve {
...
 /// Instantiate this type from the output of a digest.
 ///
-/// This can be used for implementing hash-to-scalar (e.g. as in ECDSA) or
-/// hash-to-curve algorithms.
+/// This trait is intended for use in ECDSA and should perform a conversion
+/// which is compatible with the rules for calculating `h` from `H(M)` set out
+/// in RFC6979 section 2.4. This conversion cannot fail.
+///
+/// This trait may also be useful for other hash-to-scalar or hash-to-curve
+/// use cases.
```
tarcieri, line 153: This part can probably be removed

PR #246, opened 2021-02-13
Fix compile error in ed25519 example.

newAM:
The first portion of the example in `lib.rs` does not match the second portion for `HelloVerifier` member names - and will fail to compile when combined.

This pull-request aligns the member names so the example works when the two halves are put together.


tarcieri:
Thanks!

PR #257, opened 2021-02-26
Split out RFC6979 implementation to its own crate

rvolgers:
This is the deterministic k (nonce) generation used by `ecdsa`. It is being split out because it will also be used in the upcoming `dsa` implementation.

For the currently implemented ECDSA use cases we always have a modulus which has a bit length which is divisible by 8 and equal to the length of the digest function output. This is not guaranteed to be the case for classic DSA, so some missing features had to be implemented.

(This commit contains these changes as well as the move to a separate crate. I can split those changes out again if needed, but I've been testing with the in-progress DSA code as well to make sure the interface is correct and things are working. It would take a bit of extra work and by now I just want to get something up for review.)

One of the changes was to the HMAC DRBG code, to produce output strings of arbitrary length (as defined in its standard). This was not a huge change, just adding a loop.

To properly implement the new aspects of RFC6979 the generation code needs to know the bit length of the modulus. I decided it was probably nicer to just pass the whole modulus, so the code can also check that various values are smaller than the modulus.

The modulus is only available via the `bitvec`-based `PrimeField::char_le_bits` API, and the value of scalars is also available in this way (via `to_le_bits`). So I decided to use it for all Scalar to bytes conversion. I would much prefer an API which properly defines the endianness (and I've posted a discussion of this in an issue on the traits repo), but for now this will do.

This PR depends on https://github.com/RustCrypto/traits/pull/565 (just merged) and https://github.com/RustCrypto/traits/pull/566 (not yet merged), otherwise tests will fail because of unimplemented features on the `MockCurve` used for testing.

The test suite in the `ecdsa` crate is very minimal (and does not exercise the new features needed for DSA) and I've added no dedicated tests for the new crate. However, the new DSA code has many tests based on official test vectors with many combinations of digest and key length, and these do exercise all the features except for moduli with a bit length not divisible by 8.

This is currently at a "ready for review" state and I've tested it fairly well, but I would not be surprised to find something missing a bit of polish.

### DIFF
--- a/ecdsa/Cargo.toml
+++ b/ecdsa/Cargo.toml
@@ -18,6 +18,8 @@ der = { version = "0.2", optional = true, features = ["big-uint"] }
 elliptic-curve = { version = "0.9", default-features = false }
 hmac = { version = "0.10", optional = true, default-features = false }
 signature = { version = ">= 1.3.0, < 1.4.0", default-features = false, features = ["rand-preview"] }
+rfc6979 = {path = "../rfc6979" }
+bitvec = { version = "0.20.0", default-features = false }
 
 [dev-dependencies]
 elliptic-curve = { version = "0.9", default-features = false, features = ["dev"] }
--- a/ecdsa/src/rfc6979.rs
+++ b/ecdsa/src/rfc6979.rs
@@ -4,16 +4,17 @@
 //! <https://tools.ietf.org/html/rfc6979#section-3>
 
 use crate::hazmat::FromDigest;
+use bitvec::prelude::*;
 use elliptic_curve::{
-    ff::PrimeField,
-    generic_array::GenericArray,
+    ff::{Field, PrimeField},
     ops::Invert,
     weierstrass::Curve,
     zeroize::{Zeroize, Zeroizing},
     FieldBytes, NonZeroScalar, ProjectiveArithmetic, Scalar,
 };
-use hmac::{Hmac, Mac, NewMac};
 use signature::digest::{BlockInput, FixedOutput, Reset, Update};
+
+use ::rfc6979::KGenerator;
 
 /// Generate ephemeral scalar `k` from the secret scalar and a digest of the
 /// input message.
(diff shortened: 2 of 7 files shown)


ecdsa/src/rfc6979.rs
```diff
@@ -28,82 +29,47 @@ where
...
+
+    // Verify that values roundtrip through bits_to_bytes and from_repr as expected.
+    // The from_repr interface does not guarantee big endian interpretation, and if it is not
+    // as expected the generated value may not have the required entropy.
+    let one = Scalar::<C>::one();
+    let one_bytes = bits_to_bytes(one.to_le_bits());
+    assert!(
+        one_bytes[one_bytes.len() - 1] == 1 && Scalar::<C>::from_repr(one_bytes) == Some(one),
+        "curve repr not as expected"
+    );
```
rvolgers, line 52: This is obviously kind of ugly, but I don't see how I can avoid doing something like this with the current APIs, at least not without leaving a gaping hole in the API allowing for misuse. In theory at least the compiler should be able to fully evaluate this at compile time, but I kind of doubt it's that good.

ecdsa/src/rfc6979.rs
```diff
@@ -28,82 +29,47 @@ where
...
+
+    // Convert inputs to big endian byte representation
+    let modulus = bits_to_bytes(Scalar::<C>::char_le_bits());
+    let h1 = bits_to_bytes(Scalar::<C>::from_digest(msg_digest).to_le_bits());
```
rvolgers, line 56: It might be nicer to pass the `Digest` down all the way to the new crate instead of converting everything to byte buffers here. Unfortunately the new crate can't use the `FromDigest` trait, because it's declared in `ecdsa` and depending on `ecdsa` would create a circular dependency between the `ecdsa` and `rfc6079` crates.

ecdsa/src/rfc6979.rs
```diff
@@ -28,82 +29,47 @@ where
...
+    let mut x = bits_to_bytes(secret_scalar.to_le_bits());
```
tarcieri, line 57: All of this feels like complexity that's trying to work around the potential that `to_repr` will ever be anything other than big endian.

That will *never* be the case for ECDSA, which is to say that ECDSA is undefined for any curve with a non-big-endian field element representation, and the chances of that changing at any point in the future is, for all intents and purposes, zero.

If you'd prefer to have a separate trait defined by the `ecdsa` crate which by contract will always emit a big endian-serialized `FieldBytes`, we can define that, but really I don't think there's any need for any of this, and separately there shouldn't be any need to loop in any sort of bit-level views of field elements.

ecdsa/src/sign.rs
```diff
@@ -129,7 +129,7 @@ where
     /// computed using the algorithm described in RFC 6979 (Section 3.2):
     /// <https://tools.ietf.org/html/rfc6979#section-3>
     fn try_sign_digest(&self, digest: D) -> Result<Signature<C>, Error> {
-        let k = rfc6979::generate_k(self.inner.secret_scalar(), digest.clone(), &[]);
+        let k = crate::rfc6979::generate_k(self.inner.secret_scalar(), digest.clone(), &[]);
```
rvolgers, line 132: We now have both an `rfc6979` module in `ecdsa` as well as a crate with the same name, so we need this disambiguation.

I considered moving that functionality to the new crate under an `ecdsa` feature flag or something, but as mentioned earlier I don't think that's possible without creating a circular dependency between the two crates.

rfc6979/src/hmac_drbg.rs
```diff
@@ -0,0 +1,65 @@
...
+        Self { k, v }
+    }
+
+    /// Get the next `HMAC_DRBG` output
+    pub fn generate_into(&mut self, out: &mut [u8]) {
+        for out_chunk in out.chunks_mut(self.v.len()) {
+            self.k.update(&self.v);
+            self.v = self.k.finalize_reset().into_bytes();
+            out_chunk.copy_from_slice(&self.v[..out_chunk.len()]);
+        }
```
rvolgers, line 57: Oh, I guess the "generate arbitrary amount of output" change had already landed? I guess it's not part of this PR after all.

rfc6979/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Both should have been truncated to the same number of bits as needed to represent the
+    /// modulus, interpreted as a big-endian number, reduced to being below the modulus by
+    /// subtracting the modulus if it is larger than the modulus, and then encoding as a big
+    /// endian number, padded with leading zero bytes to the same length as the modulus.
+    /// See RFC6979 for details.
+    ///
+    /// The modulus must be in big endian format with no padding (i.e. first byte must not be 0).
+    ///
+    /// This function will panic if these requirements are not met.
+    pub fn new<T: AsRef<[u8]>>(modulus: I, secret: T, digest: T, additional_data: &[u8]) -> Self {
```
rvolgers, line 42: Separate generics `I` and `T` is a bit ugly here, I think the idea was to allow user to pick an owned or borrowed representation for the modulus?

rfc6979/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+        let rng = {
+            let modulus = modulus.as_ref();
+            let secret = secret.as_ref();
+            let digest = digest.as_ref();
+            assert!(modulus.len() > 0 && modulus[0] != 0);
+            assert!(secret.len() == modulus.len() && secret < modulus);
+            assert!(digest.len() == modulus.len() && digest < modulus);
```
rvolgers, line 49: This pattern is also used in `generate_into`: if the length of two big-endian byte buffers is equal, doing a numeric comparison is equivalent to lexical comparison between the buffers.

rvolgers (commented):
Added some notes

ecdsa/Cargo.toml
```diff
@@ -18,6 +18,8 @@ der = { version = "0.2", optional = true, features = ["big-uint"] }
 elliptic-curve = { version = "0.9", default-features = false }
 hmac = { version = "0.10", optional = true, default-features = false }
 signature = { version = ">= 1.3.0, < 1.4.0", default-features = false, features = ["rand-preview"] }
+rfc6979 = {path = "../rfc6979" }
+bitvec = { version = "0.20.0", default-features = false }
```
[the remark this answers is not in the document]
tarcieri: Per https://github.com/zkcrypto/ff/pull/42 I think it'd be nice if we could make the `bitvec` dependency optional and if that's possible I would like for it not to be required for ECDSA

ecdsa/src/rfc6979.rs
tarcieri: If it's helpful to push the `FromDigest` trait down to the `rfc6979` crate, I guess I'd be ok with having it there.

[As I mentioned earlier](https://github.com/RustCrypto/signatures/pull/244#issuecomment-778550307) in ECDSA the specific transformation is a core part of the algorithm and not specifically related to RFC6979, but what ECDSA does and what RFC6979 needs are the same, so I guess I wouldn't be opposed to moving it.

ecdsa/src/sign.rs
tarcieri: That's fine for now. I can look into refactoring later.

rfc6979/src/hmac_drbg.rs
tarcieri: Yes I merged it earlier (#243)

ecdsa/src/rfc6979.rs
rvolgers: I agree that ECDSA with the wrong kind of representation does not make sense. I just want to make sure it explodes instead of silently doing the wrong thing.

I also agree on not wanting to deal with `bitvec`. The endian thing is not the only reason why I need it though! The `char_le_bits` API is the only way to get the modulus of the prime field, and it returns a `bitvec` based type too.

rvolgers: Also, worth pointing out that there is already an ECDSA specific trait: `FromDigest`. Perhaps it would be better to turn it into an `ECDSACompatible` trait or something, and add "must have a big endian repr" to its contract? I kind of hate the name, but just to get the idea across.

ecdsa/Cargo.toml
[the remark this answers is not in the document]
rvolgers: I also think this would be nice.

ecdsa/src/rfc6979.rs
tarcieri: > I agree that ECDSA with the wrong kind of representation does not make sense. I just want to make sure it explodes instead of silently doing the wrong thing.

The requirements for integrating the `ecdsa` crate with a particular curve implementation are a pretty tall order right now. We made the API deliberately coarse grained both to allow hiding certain things from the public API (i.e. base field elements) but also permitting optimizations that are only possible by leveraging potentially unique features of curve arithmetic backends.

- https://docs.rs/ecdsa/0.10.2/ecdsa/hazmat/trait.SignPrimitive.html
- https://docs.rs/ecdsa/0.10.2/ecdsa/hazmat/trait.VerifyPrimitive.html

Concrete implementations end up looking like this:

- https://github.com/RustCrypto/elliptic-curves/blob/master/k256/src/ecdsa/sign.rs#L211-L237
- https://github.com/RustCrypto/elliptic-curves/blob/master/k256/src/ecdsa/verify.rs#L110-L130

All of these things are explicitly called out as `hazmat` in the API and the bar is pretty high for even getting a particular implementation working in the first place, so I'm not too particularly worried that someone will manage to produce an ECDSA implementation which is otherwise working but manages to get everything else but endianness wrong.

rvolgers: Ah, that makes a lot of sense. I had seen those traits briefly because I am just starting to think about what kind of public interface I want for the `dsa` code (it started from a very different place, with everything dynamic in the `rpgp` style instead of the mostly static dispatch style more common in RustCrypto), but had not really gone through them or considered why they are the way they are. This is good background information also for that work.

I'll take some time to digest this and then see how far I get making something we're both happy with.

tarcieri:
Closing in favor of #409

PR #321, opened 2021-06-18
Add zeroization on drop for SigningKey

fjarri:
I am not exactly sure what is the convention for Drop implementations, but it seems like `SigningKey` needs one.

### DIFF
--- a/ecdsa/src/sign.rs
+++ b/ecdsa/src/sign.rs
@@ -86,6 +86,17 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Thanks. I think this was overlooked as part of #302.

I hope we can get back to `SigningKey` being a newtype for `SecretKey` so we can ensure all of these sorts of concerns are handled there, but that will require a generic `Scalar` type which works for all curves.

tarcieri:
Released in v0.12.2

PR #337, opened 2021-07-21
feat(ed255519): Introduce `serde_bytes` optional dependency

connec:
This allows consumers to opt-in to `serde_bytes` support by setting the
`std`, `serde`, and `serde_bytes` features. This makes it convenient to
optimise (de)serialization for formats with efficient byte array
encodings, e.g.

    #[derive(serde::Deserialize, serde::Serialize)]
    struct MySignature {
        #[serde(with = "serde_bytes")]
        signature: ed25519::Signature,
    }

Without first-class support for `serde_bytes`, consumers would have to
write their own (de)serialization machinery for this common case.

### DIFF
--- a/ed25519/src/lib.rs
+++ b/ed25519/src/lib.rs
@@ -256,8 +256,10 @@
 
 #[cfg(feature = "serde")]
 use serde::{de, ser, Deserialize, Serialize};
+#[cfg(feature = "serde_bytes")]
+use serde_bytes_crate as serde_bytes;
 
-#[cfg(all(test, feature = "std"))]
+#[cfg(all(feature = "std", any(test, feature = "serde_bytes")))]
 extern crate std;
 
 pub use signature::{self, Error};
@@ -402,6 +404,49 @@ impl<'de> Deserialize<'de> for Signature {
     }
 }
 
+#[cfg(feature = "serde_bytes")]
+impl serde_bytes::Serialize for Signature {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_bytes(&self.0)
+    }
+}
+
+#[cfg(feature = "serde_bytes")]
+impl<'de> serde_bytes::Deserialize<'de> for Signature {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        struct ByteArrayVisitor;
+
+        impl<'de> de::Visitor<'de> for ByteArrayVisitor {
+            type Value = [u8; SIGNATURE_LENGTH];
+
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("bytestring of length 64")
+            }
+
+            fn visit_bytes<E>(self, bytes: &[u8]) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                use de::Error;
+
+                bytes
+                    .try_into()
+                    .map_err(|_| Error::invalid_length(bytes.len(), &self))
+            }
+        }
+
+        deserializer
+            .deserialize_bytes(ByteArrayVisitor)
+            .map(Signature::from)
+    }
+}
+
 #[cfg(all(test, feature = "serde", feature = "std"))]
 mod tests {
     use super::*;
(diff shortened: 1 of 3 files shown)


connec:
In contrast to #338 this has the advantage of keeping the features orthogonal, so `--all-features` still works fine and testing is kept simple. The downside is an additional dependency in the tree, and more complex integration for consumers (e.g. the need for an extra attr).

ed25519/src/lib.rs
```diff
@@ -254,10 +254,16 @@
 #![forbid(unsafe_code)]
 #![warn(missing_docs, rust_2018_idioms, unused_qualifications)]
 
+#[cfg(all(
+    feature = "serde_bytes",
+    any(not(feature = "serde"), not(feature = "std"))
+))]
+compile_error!("Feature \"serde_bytes\" requires \"std\" and \"serde\" features to be enabled");
```
tarcieri, line 261: It seems like it should be possible to make this "just work", although if you'd prefer I can tackle that as a followup.

tarcieri:
@connec I think overall I prefer this approach to one which introduces crate features which don't properly unify (i.e. #338 

It does introduce an extra dependency, but I'm okay with that given the provenance of that particular dependency (i.e. it's by `serde`'s authors and therefore is as trustworthy as `serde` itself)

ed25519/src/lib.rs
```diff
@@ -402,6 +408,62 @@ impl<'de> Deserialize<'de> for Signature {
...
+            fn visit_byte_buf<E>(self, bytes: std::vec::Vec<u8>) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                use de::Error;
+
+                let bytes_len = bytes.len();
+                bytes
+                    .try_into()
+                    .map_err(|_| Error::invalid_length(bytes_len, &self))
```
tarcieri, line 457: I think this might require bumping MSRV to 1.47, which is fine

ed25519/src/lib.rs
connec: I did spend some time playing with it, but wasn't able to make it work. I felt blocked by the limitation against having features in `Cargo.toml` with the same name as an optional dependency. I did use a separate feature at one point, like:

```toml
[features]
serde_bytes_support = ["serde", "serde_bytes", "std"]
```

This works, but I felt it wasn't much better since it doesn't remove the implicit `serde_bytes` feature from the optional dependency (so you could still compile with just `--features serde_bytes` and get either missing type errors or silent nothing).

ed25519/src/lib.rs
connec: Ah yeah, looks like `TryFrom<Vec<T>>` was stabilised in 1.48: https://doc.rust-lang.org/stable/std/primitive.array.html#impl-TryFrom%3CVec%3CT%2C%20A%3E%3E. I can use a different approach, probably not worth bumping MSRV just for this.

ed25519/src/lib.rs
tarcieri: You can prevent a clash between the crate name and the feature name by renaming the crate on import:

```toml
serde_bytes_crate = { package = "serde_bytes", version = "0.11", optional = true }
```

connec: Cool, that would allow for something like:

```toml
[features]
serde_bytes = ["serde", "serde_bytes_crate", "std"]
```

I suppose that would leave an implicit `serde_bytes_crate` feature, which is also a bit annoying, but if it's preferable to the `compile_error!` I can make that change.

tarcieri: That can eventually be sorted out / cleaned up when namespaced features land

ed25519/src/lib.rs
```diff
@@ -402,6 +404,49 @@ impl<'de> Deserialize<'de> for Signature {
     }
 }
 
+#[cfg(all(feature = "serde", feature = "serde_bytes"))]
```
tarcieri, line 407: Minor nit, but these are redundant as `serde_bytes` implies `serde`

```suggestion
#[cfg(feature = "serde_bytes")]
```

connec:
I've pushed an update that removes the `compile_error!` and the dependence on `TryFrom<Vec<T>> for [T; N]`. For the latter, based on [the docs](https://docs.rs/serde/1.0.126/serde/de/trait.Visitor.html#method.visit_byte_buf) I've just removed the `visit_byte_buf` method, since there's no way to create a >32 element array from a `Vec` with a copy as far as I can tell.

tarcieri:
Looks good, thank you!

It might be good to add the example from the description to the rustdoc as well.

tarcieri:
Released in #340

PR #338, opened 2021-07-21
feat(ed25519): Add a feature to serialize with serde's bytes type

### LINKED ISSUE #336: ed25519: Sub-optimal serialization for type-tagged formats
As of #78, the encoding for formats that "serialize byte slices more compactly or more efficiently than other types of slices" (to quote [serde's docs](https://docs.rs/serde/1.0.126/serde/trait.Deserializer.html#tymethod.deserialize_bytes)) may be worse.

For example, using Message Pack (via [`rmp_serde`](https://docs.rs/rmp-serde/)), the tuple-based approach leads to encoding as an array where each element is individually encoded, hence each element needs an additional type tag byte (though Message Pack encodes integers <128 with a single byte, so on average only ~50% of values get inflated this way).

Although my specific use-case is with Message Pack, I assume this issue would affect other formats with specific byte array encodings, though I don't know how common that is (as serde has first-class support in their data model, I'd guess it's not too uncommon?).

It's certainly possible to work around this downstream using `#[serde(with = "...")]` et-al, but I wanted to check whether this was considered when implementing #78, and considered desired behaviour, or if the bincode optimisation could be dropped for broader compatibility.

*Edit:* I guess this is basically a choice between known-size or known-type optimisations. A known-size byte array could benefit from both, but serde's data model doesn't support it.

connec:
An alternative fix for #336 introducing a feature to swap between `*_bytes` and non-`*_bytes` serialization – the two are mutually exclusive!

---

This introduces a new feature, `serde_as_bytes`, that switches the
implementation of (de)serialization to use serde's `*_bytes` methods,
allowing crate consumers to opt-in to this approach which is beneficial
for formats with optimised byte array representations.

This has required a small change to the test workflow, since it's
necessary to test both with and without the `serde_as_bytes` feature.
The features cannot be made mutually exclusive, though logically they
are, without changing the `serde` feature flag, which would be
backwards-incompatible.

Fixes #336.

### DIFF
--- a/.github/workflows/ed25519.yml
+++ b/.github/workflows/ed25519.yml
@@ -101,5 +101,8 @@ jobs:
       - name: Run cargo test --lib
         run: cargo test --lib --release
 
-      - name: Run cargo test --all-features
-        run: cargo test --all-features --release
+      - name: Run cargo test --features serde
+        run: cargo test --features serde --release
+
+      - name: Run cargo test --features serde_as_bytes
+        run: cargo test --features serde_as_bytes --release
--- a/ed25519/Cargo.toml
+++ b/ed25519/Cargo.toml
@@ -23,4 +23,5 @@ rand_core = { version = "0.5", features = ["std"] }
 
 [features]
 default = ["std"]
+serde_as_bytes = ["serde"]
 std = ["signature/std"]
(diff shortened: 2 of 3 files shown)


connec:
Closing in favour of #337.

PR #352, opened 2021-09-02
ecdsa: bump `elliptic-curve` to v0.11.0-pre

tarcieri:
Sourced via git.

This includes transitive updates to `ff` and `group` v0.11.

### DIFF
--- a/ecdsa/src/lib.rs
+++ b/ecdsa/src/lib.rs
@@ -98,7 +98,7 @@ use core::{
 };
 use elliptic_curve::{
     bigint::Encoding as _,
-    generic_array::{sequence::Concat, typenum::Unsigned, ArrayLength, GenericArray},
+    generic_array::{sequence::Concat, ArrayLength, GenericArray},
     FieldBytes, FieldSize, ScalarBytes,
 };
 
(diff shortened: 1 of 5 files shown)


ecdsa/src/lib.rs
```diff
@@ -176,38 +186,36 @@ where
...
-                was_high
-            })
-            .ok_or_else(Error::new)
+        let (s_low, was_high) = self.s().normalize_low();
+
+        if was_high {
+            self.bytes[C::UInt::BYTE_SIZE..].copy_from_slice(&s_low.to_repr());
+        }
+
+        was_high
```
webmaster128, line 218: Do I understand this diff correctly that `normalize_s` cannnot error anymore? Was there a possible error case before or did we use the `Result` only to make the type system happy?

tarcieri: Previously `ecdsa::Signature` was a naive wrapper around a 64-byte array which could contain any value.

[The bytes are now validated by the constructor](https://github.com/RustCrypto/signatures/blob/6ddc647172c76b4ead0b2ed4d2e37c64812348e9/ecdsa/src/lib.rs#L280-L304) and guaranteed to always represent the big endian serialization of a valid nonzero element of the scalar field. Even better: this now works for any elliptic curve regardless of if it has a field arithmetic implementation by using a common implementation in the `elliptic-curve` crate.

That eliminates the error case that previously existed here, where it was possible that the bytes would not decode successfully as `NonZeroScalar`.

webmaster128: Great, thanks a lot for the detailed answer. Much appreciated.

PR #433, opened 2022-01-17
Update elliptic-curve and digest

daxpedda:
This builds on RustCrypto/traits#883.

~~The changes are minimal,~~ a lot of added where bounds because of `Hmac` and changed `Invert`s associated type `Output` to `CtOption<T>`.

<details><summary><strike>To separate or not to separate</strike></summary>
Currently there is an issue with having ring-compat reference a different version of signature then what the workspace has:
```
error: failed to select a version for `signature`.
    ... required by package `ecdsa v0.13.3`
    ... which satisfies dependency `ecdsa = "^0.13"` of package `ring-compat v0.4.0`
    ... which satisfies dependency `ring-compat = "^0.4"` of package `ed25519 v1.4.0-pre (**/ed25519)`
versions that meet the requirements `>=1.3.1, <1.5` are: 1.4.0, 1.3.2, 1.3.1

all possible versions conflict with previously selected packages.

  previously selected package `signature v1.5.0`
    ... which satisfies dependency `signature = "^1.5"` (locked to 1.5.0) of package `ecdsa v0.13.4 (**/ecdsa)`

failed to select a version for `signature` which could resolve this conflict
```
I circumvented this by updating ring-compat, but that is something I don't want to do because I have no clue what I'm doing there:
https://github.com/RustCrypto/ring-compat/compare/master...khonsulabs:ecdsa-digest

So I say either I actually manage to update rust-compat with some feedback, or we find a way to resolve this, for example by removing ed25519 from the workspace. Like is done in the traits repository with elliptic-curve. I would prefer not updating rust-compat.
</details>
I decided to go ahead and exclude ed25519 from the workspace. This should make updating easier.


ecdsa/Cargo.toml
```diff
@@ -15,17 +15,17 @@ edition = "2021"
 rust-version = "1.56"
 
 [dependencies]
-elliptic-curve = { version = "0.11.6", default-features = false, features = ["sec1"] }
-signature = { version = ">= 1.3.1, <1.5", default-features = false, features = ["rand-preview"] }
+elliptic-curve = { git = "https://github.com/khonsulabs/traits", branch = "digest", default-features = false, features = ["sec1"] }
```
tarcieri, line 18: I'd suggest using a `patch.crates-io` directive for this so you don't need to duplicate the git URL everywhere the crate is referenced

tarcieri:
@daxpedda you should be able to update this to use https://github.com/RustCrypto/traits.git as the upstream repo, and pin the `elliptic-curve` version to `v0.12.0-pre`

ecdsa/Cargo.toml
daxpedda: The reason I did it like this is to make it easy for current consumers of this branch to use it without having to spam `patch`es throughout their crates. I am assuming this is no problem because it's not intended to merge this without git dependency anyway?

daxpedda:
Done.

ecdsa/Cargo.toml
```diff
@@ -15,17 +15,17 @@ edition = "2021"
...
 [dependencies]
-elliptic-curve = { version = "0.11.6", default-features = false, features = ["sec1"] }
-signature = { version = ">= 1.3.1, <1.5", default-features = false, features = ["rand-preview"] }
+elliptic-curve = { git = "https://github.com/RustCrypto/traits", version = "0.12.0-pre", default-features = false, features = ["sec1"] }
+signature = { version = "1.5", default-features = false, features = ["rand-preview"] }
 
 # optional dependencies
 der = { version = "0.5", optional = true }
-rfc6979 = { version = "0.1", optional = true, path = "../rfc6979" }
+rfc6979 = { git = "https://github.com/khonsulabs/signatures", branch = "elliptic-curve-digest", optional = true }
```
daxpedda, line 23:
```suggestion
rfc6979 = { version = "0.1", optional = true, path = "../rfc6979" }
```
This needs to be done when this is ready to be merged.

ecdsa/Cargo.toml
tarcieri: I've done both `git`-based dependencies and prereleases for `elliptic-curve` in the past when a new version is under development.

If you'd like me to cut a prerelease I can, but `git` is a little more agile for development purposes.

ecdsa/Cargo.toml
tarcieri: Not sure why you had to change this at all?

daxpedda: Same reason as above, so I can depend on it without `[patch]`.

tarcieri: But it's a crate in the same repository referred to by `path` already?

tarcieri: Anyway I can't merge this until you change it back...

daxpedda: Yes, but when you depend on `ecdsa` as a git dependency it will still pull `rfc6979` in from crates.io. I think?

tarcieri: And this is where `[patch.crates-io]` is helpful.

Can you change it back and then figure out what to do downstream?

ecdsa/Cargo.toml
daxpedda: If cutting a pre-release for elliptic-curve is doable then that would be great, it would unblock our projects from a release on crates.io too.

Otherwise if you want to merge this it's no problem for me to remove this and replace it with `[patch]`.

ecdsa/Cargo.toml
daxpedda: I can do that, no problem.

ecdsa/Cargo.toml
tarcieri: I can do that

ecdsa/Cargo.toml
daxpedda: Done.

tarcieri:
`elliptic-curve` v0.12.0-pre.0 has been released to crates.io.

Note that MSRV is now 1.57.

daxpedda:
The CI had to change a bit because I excluded ed25519 from the workspace, I hope that's alright.

tarcieri:
Why did you need to exclude it from the workspace?

All three crates in the workspace are 2021 edition crates.

tarcieri:
Aha, makes sense

PR #439, opened 2022-01-19
Use `from_be_digest_reduced`

daxpedda:
Builds on RustCrypto/traits#904.

Does this look about right or is this overkill?


tarcieri (approved):
Looks good. Want to mark it as ready?

PR #441, opened 2022-01-19
Fix broken workspace jobs

### LINKED ISSUE #440: Broken CI because of `working-directory`
It seems I have broken the CI because I didn't know that you can only use `working-directory` with `run` and not with `uses`. The reason I didn't notice was because for some reason these jobs didn't show up in the PR.

daxpedda:
Fixes #440.

This also adds updating ed25519 by Dependabot.


tarcieri:
I did a simpler fix in #442

PR #447, opened 2022-02-12
Add license files

dimbleby:
It's conventional, and helpful for tooling, to include copies of licenses in your project.  The README is clear that this project is licensed under either of MIT or Apache licenses, so those are the two I've added.

In the copyright statements I've used "RustCrypto/signatures authors" and "2018" (which seems to be when the first commit was).

If you think this is a sensible idea but have a healthy paranoia about whether strangers on the internet can be trusted to provide license files, you might prefer to reject this MR and achieve the same end by using the github helper https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository

### DIFF
--- a/ecdsa/LICENSE-APACHE
+++ b/ecdsa/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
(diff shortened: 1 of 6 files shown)


dimbleby:
Yeah, for consistency with the rest of the project I've moved licenses into the individual crates, and assigned copyright to "RustCrypto Developers"

tarcieri:
Thank you!

PR #471, opened 2022-05-09
dsa: Add initial DSA implementation

aumetra:
This PR adds an initial implementation of DSA (see #8)  

The following things work when tested against OpenSSL:

- The generated keys are valid and can be imported and exported from/to their DER/PEM representation
- Signatures generated by this library can be successfully verified
- Signatures can be imported and exported from/into their DER representation
- Signatures generated by OpenSSL can be successfully imported and verified

Things that need to be looked at:

- Additional test coverage (~~especially verification of OpenSSL generated signatures~~ (edit: added with commit 40e7fd2))
- Code structure (is the current file structure fine or should it be changed?)
- ~~`signature` compatibility (the crate itself isn't deeply integrated with `signature` and only offers compatibility wrappers (see the `compat.rs` file) Should those traits be integrated more directly?)~~ (edit: The `signature` crate has been integrated more tightly with commit 7123e0f)


tarcieri:
Thank you for contributing this!

I'll try to do a more thorough review when I have some time. Just glancing through quickly I don't see any immediate concerns.

aumetra:
@tarcieri No problem, take your time! 

Out of interest, would it be okay to use one of the `signature` preview APIs? In particular `RandomizedDigestSigner`.  
This trait would fit pretty well for the current implementation and would allow us to get rid of most compatibility wrappers

aumetra:
@tarcieri No problem, take your time! 

Out of interest, would it be okay to use one of the `signature` preview APIs? In particular `RandomizedDigestSigner`.  
This trait would fit pretty well for the current implementation and would allow us to get rid of most compatibility wrappers

tarcieri:
Yes, it's fine to use the `digest` feature of `signature`. Just make sure to lock to the minor version.

The `ecdsa` crate also does this if you'd like to use it as an example.

tarcieri:
One thing I thought I might mention before I can do a more thorough review is RFC6979 deterministic signature support.

The `rfc6979` crate is designed with the intention of being potentially applicable to DSA as well as ECDSA, although some breaking changes might be needed. Fortunately, it's not a part of the `ecdsa` crate's public API so we're free to make such breaking changes, it's just the `ecdsa` crate will need to be updated too.

[`HmacDrbg`](https://docs.rs/rfc6979/latest/rfc6979/struct.HmacDrbg.html) is part of its public API so you're also free to just use that if not the [`generate_k`](https://docs.rs/rfc6979/latest/rfc6979/fn.generate_k.html) function which is presently coupled to `crypto-bigint`.

aumetra:
The test for the deterministic signatures is pretty messy and doesn't include all test vectors but should be good enough for a start

dsa/README.md
```diff
@@ -0,0 +1,52 @@
+# dsa
```
tarcieri, line 1: Can you use one of our existing README.mds as a template? We try to keep them fairly consistent across the project:

https://github.com/RustCrypto/signatures/blob/master/ecdsa/README.md

Thanks!

aumetra: I hope the about section in the new README is good enough. I wasn't exactly sure what to write there

dsa/tests/privatekey.rs
```diff
@@ -25,7 +25,7 @@ fn decode_encode_openssl_private_key() {
     assert!(private_key.is_valid());
 
     let reencoded_private_key = private_key
-        .to_pkcs8_pem(LineEnding::LF)
+        .to_pkcs8_pem(LineEnding::default())
```
tarcieri, line 28: I'd suggest keeping these on `LineEnding::LF` so the tests pass consistently regardless of operating system

aumetra: There was this CI failure [here](https://github.com/RustCrypto/signatures/runs/6434988350?check_suite_focus=true) which failed because the OpenSSL PEM file had CRLF line endings (the checkout action probably changed the line endings). That's why I adjusted it to using `LineEnding::default()` 

But yes, I should probably adjust the CI instead of changing the tests

dsa/src/lib.rs
```diff
@@ -0,0 +1,69 @@
+#![no_std]
+#![forbid(missing_docs, unsafe_code)]
+#![deny(rust_2018_idioms)]
```
tarcieri, line 3: We generally use `warn` rather than `deny`, and enforce code is warning-free in CI

```suggestion
#![forbid(unsafe_code)]
#![warn(missing_docs, rust_2018_idioms)]
```

(`unsafe_code` is the one exception in crates where we don't want any)

dsa/src/privatekey.rs
```diff
@@ -0,0 +1,200 @@
+//!
```
tarcieri, line 1: Nit: we generally call these modules `private_key.rs` which follows a camel case -> snake case conversion

dsa/src/publickey.rs
```diff
@@ -0,0 +1,145 @@
+//!
```
tarcieri, line 1: Nit: would suggest naming this file `public_key.rs`

tarcieri (changes requested):
This is looking pretty close to an MVP. Just a few nits.

tarcieri (approved):
Thank you!

PR #484, opened 2022-05-16
dsa: description and boilerplate fixups

tarcieri:
- add description to Cargo.toml
- minor DER signature fixup
- set `opt-level = 2` at workspace level so tests run fast by default


aumetra:
Should the dsa `Cargo.toml` have the `repository` field set, too?

PR #489, opened 2022-05-20
dsa: Small API overhaul

aumetra:
This PR overhauls the API of the `dsa` crate a bit 

# Changes 

(copy pasted from `CHANGELOG.md`)

## Added

- Internal sanity check validating the `r` and `s` components of the signature
- Public `OID` constant representing the object identifier defined in RFC-3279, section 2.3.2 

## Changed

- `Components::generate` now takes an `KeySize` struct instead of an `(u32, u32)` tuple
- `Components::from_components`, `SigningKey::from_components` and `VerifyingKey::from_components` are now fallible and validate themselves upon creation

## Removed

- Removed `is_valid` functions on `Components`, `SigningKey` and `VerifyingKey` (successful construction/deserialisation now implies validity)


dsa/src/components.rs
```diff
@@ -63,7 +68,7 @@ impl Components {
 
     /// Check whether the components are valid
     #[must_use]
-    pub fn is_valid(&self) -> bool {
+    pub(crate) fn is_valid(&self) -> bool {
```
tarcieri, line 71: Perhaps just factor this method into the constructor?

dsa/src/components.rs
```diff
@@ -81,7 +86,9 @@ impl<'a> DecodeValue<'a> for Components {
         let q = BigUint::from_bytes_be(q.as_bytes());
         let g = BigUint::from_bytes_be(g.as_bytes());
 
-        Ok(Self::from_components(p, q, g))
+        // TODO: Using this error doesn't seem right
```
tarcieri, line 89: This is the correct error to use for the DER deserializer. You can go ahead and remove this comment.

dsa/src/components.rs
```diff
@@ -81,7 +86,9 @@ impl<'a> DecodeValue<'a> for Components {
...
+        Self::from_components(p, q, g)
+            .map_err(|_| der::Error::new(ErrorKind::Value { tag: Tag::Integer }, reader.position()))
```
tarcieri, line 91: This can more succinctly be written as:

```suggestion
            .map_err(|_| reader.error(Tag::Integer.value_error()))
```

dsa/src/generate/keypair.rs
```diff
@@ -16,6 +16,8 @@ where
     let x = rng.gen_biguint_range(&BigUint::one(), components.q());
     let y = components::public(&components, &x);
 
-    let verifying_key = VerifyingKey::from_components(components, y);
+    let verifying_key = VerifyingKey::from_components(components, y)
+        .expect("[Bug] Newly generated verifying key considered invalid");
     SigningKey::from_components(verifying_key, x)
+        .expect("[Bug] Newly generated signing key considered invalid")
```
tarcieri, line 22: Can you perhaps chain this using `and_then` so you don't have to duplicate the `expect`?

dsa/src/lib.rs
```diff
@@ -27,40 +29,44 @@
...
 
 mod components;
 mod generate;
 mod sig;
 mod signing_key;
+mod size;
 mod verifying_key;
 
+/// DSA object identifier as defined by RFC-3279, section 2.3.2
+const DSA_OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.10040.4.1");
```
tarcieri, line 68: How about:

```suggestion
pub const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.10040.4.1");
```

...accessible as `dsa::OID`?

dsa/src/components.rs
aumetra: Thanks for noticing this. I moved all the validation checks into the respective constructors

dsa/src/lib.rs
aumetra: Sounds like a nice addition. Added

dsa/src/size.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// Bit size of q
+    pub(crate) n: u32,
+}
+
+impl KeySize {
+    define_param_size!(deprecated: 1024, 160);
+    define_param_size!(2048, 224);
+    define_param_size!(2048, 256);
+    define_param_size!(3072, 256);
```
tarcieri, line 30: It feels to me like these macros are just adding indirection and an additional dependency with `paste`.

Can't this simply be written as:

```suggestion
    /// DSA parameter size constant: L = 1024, N = 160
    #[deprecated(note = "This size constant has a security strength of under 112 bits per SP 800-57 Part 1 Rev. 5")]
    pub const DSA_1024_160: Self = Self { l: 1024, n: 160 };
    
    /// DSA parameter size constant: L = 2048, N = 224
    pub const DSA_2048_224: Self = Self { l: 2048, n: 224 };

    /// DSA parameter size constant: L = 2048, N = 256
    pub const DSA_2048_256: Self = Self { l: 2048, n: 256 };

    /// DSA parameter size constant: L = 3072, N = 256
    pub const DSA_3072_256: Self = Self { l: 3072, n: 256 };
```

aumetra: When I originally wrote the code I didn't like the idea of copy pasting a bunch of code and the dependency tree already had `paste` in it so I wrote this macro  

But here it's just an unnecessary dependency + indirection, I agree

dsa/src/size.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// DSA parameter size constant: L = 3072, N = 256
+    pub const DSA_3072_256: Self = Self { l: 3072, n: 256 };
+
+    /// Define a custom parameter size
+    ///
+    /// **⚠ YOU MOST LIKELY DO NOT NEED TO AND SHOULD NOT USE THIS!**
+    pub const fn custom(l: u32, n: u32) -> Self {
+        Self { l, n }
+    }
```
tarcieri, line 31: One nit/sidebar looking at this:

It doesn't maintain any kind of invariants, which makes it somewhat pointless unless the goal is to potentially add invariants in the future.

Invariants could be maintained simply by using `assert!`, which will const panic at compile time.

I'd suggest either adding some invariants, or making `l` and `n` into `pub` fields so they can be set directly.

tarcieri (approved):
This looks ready to go. Can you remove draft?

dsa/src/size.rs
aumetra: That was actually one of the things I wanted to talk about.  

Does it make sense to keep a functionality for defining custom sizes or should we just say "NIST recommends these sizes. Choose one of them"?  

Because those the non-standard parameters could still be generated by an external tool and then imported and used with the `dsa` crate to generate key pairs if having non-standard parameter sizes is that important

tarcieri: We could remove it for now and see if a feature request for it comes up

PR #508, opened 2022-08-01
Tighten `signature` version requirement of `ecdsa` crate

### LINKED ISSUE #507: `signature` version requirement of the `ecdsa` crate
Isn't the `signature` version requirement of `ecdsa` too lax to comply with semver?  

The `signature` crate reserves itself the right to bump the MSRV and the `digest` and `rand` dependencies in a minor version bump.  
Currently `ecdsa` has the version requirement of `^1.5` which includes `1.6`, `1.7`, etc. which could have breaking changes regarding either the MSRV or the above mentioned dependencies.

aumetra:
Tightens the version requirement of the `signature` crate to `>= 1.5, < 1.6` (same as the `dsa` crate) to avoid a potential semver breakage.

Closes #507

### DIFF
--- a/ecdsa/Cargo.toml
+++ b/ecdsa/Cargo.toml
@@ -17,7 +17,7 @@ rust-version = "1.57"
 
 [dependencies]
 elliptic-curve = { version = "0.12", default-features = false, features = ["digest", "sec1"] }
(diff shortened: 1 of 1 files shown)


tarcieri:
Thanks!

PR #509, opened 2022-08-02
ecdsa: add `ZeroizeOnDrop` marker for `SigningKey`

fjarri:
`ecdsa` depends on `elliptic-curve` 0.12, which depends on `zeroize` 1.5, so the trait should be safe to use.


tarcieri:
Thank you!

PR #511, opened 2022-08-13
rsa: add crate with generic RSA signature traits

lumag:
Add crate declaring generic RSA traits used for PKCS#1 v1.5 and PSS
signatures.

Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>

### DIFF
--- a/rsa/src/sign.rs
+++ b/rsa/src/sign.rs
@@ -1,0 +1,15 @@
+//! RSA signing key.
+
+use crate::Signature;
+use digest::DynDigest;
+use signature::{RandomizedSigner, Signer};
+
+/// RSA Signing key using PKCS1v15 padding
+pub trait PKCS1v15SigningKey: Signer<Signature> {}
+/// RSA Verifying key using PKCS1v15 padding
+pub trait PKCS1v15BlindedSigningKey: RandomizedSigner<Signature> {}
+
+/// RSA Signing key using PSS padding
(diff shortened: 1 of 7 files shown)


tarcieri:
I'd suggest keeping things in the `rsa` crate now. Something like this might eventually make sense, but I think as an extraction as a working implementation from the `rsa` crate.

I'll leave a few other notes.

rsa/src/sign.rs
```diff
@@ -0,0 +1,15 @@
...
+
+/// RSA Signing key using PKCS1v15 padding
+pub trait PKCS1v15SigningKey: Signer<Signature> {}
+/// RSA Verifying key using PKCS1v15 padding
+pub trait PKCS1v15BlindedSigningKey: RandomizedSigner<Signature> {}
+
+/// RSA Signing key using PSS padding
+pub trait PSSSigningKey<D: DynDigest>: RandomizedSigner<Signature> {}
+/// RSA Verifying key using PSS padding
+pub trait PSSBlindedSigningKey<D: DynDigest>: RandomizedSigner<Signature> {}
```
tarcieri, line 15: I'd suggest having separate `pkcs1v15::Signature` and `pss::Signature` types. Then you don't need separate traits like this. Instead they're just `Signer<pkcs1v15::Signature>` versus `Signer<pss::Signature>`.

The compiler can even infer which impl to use from the return type.

lumag:
Closing in favour of moving the `Signature` defs into the RSA crate itself.

PR #513, opened 2022-08-15
dsa: Allow signature v1.6.x

aumetra:
Adjust the `signature` version requirement of the `dsa` crate to include v1.6.x.

`digest`, `rand_core`, as well as the MSRV have remained compatible.


tarcieri:
Thank you!

PR #524, opened 2022-09-05
ed25519: Fix compilation of `pem` feature

aumetra:
The `ed25519` crate fails to compile with only the `pem` feature flag since the current implementation depends on the `alloc` crate (via `alloc::string::{String, ToString}`).

The `pkcs8`'s `pem` feature already depends on the `alloc` crate anyway, so enabling the `ed25519`'s `alloc` feature together with the `pem` feature should be fine.

### DIFF
--- a/ed25519/Cargo.toml
+++ b/ed25519/Cargo.toml
@@ -35,7 +35,7 @@ rand_core = { version = "0.5", features = ["std"] }
 [features]
 default = ["std"]
 alloc = ["pkcs8/alloc"]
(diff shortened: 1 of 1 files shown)


tarcieri:
Thanks! Can you add a check it builds with the `pem` feature alone to the CI config here:

https://github.com/RustCrypto/signatures/blob/master/.github/workflows/ed25519.yml#L36-L39

tarcieri:
Actually I'll take care of that as a follow-up

PR #534, opened 2022-09-12
ecdsa: impl `signature::hazmat::{PrehashSigner, PrehashVerifier}`

tarcieri:
Adds "hazmat" trait impls to `SigningKey` and `VerifyingKey` respectively which allow computing signatures over raw message digests. See RustCrypto/traits#1099.

The implementation allows digests which are shorter or longer than the field size of the curve, using zero-padding if the digest is too short, and truncating if it's too long. The minimum digest size is set to half of the curve's field size.

### DIFF
--- a/ecdsa/src/hazmat.rs
+++ b/ecdsa/src/hazmat.rs
@@ -12,27 +12,28 @@
 
 #[cfg(feature = "arithmetic")]
 use {
-    crate::{Error, RecoveryId, Result, SignatureSize},
+    crate::{RecoveryId, SignatureSize},
     core::borrow::Borrow,
     elliptic_curve::{
         group::Curve as _,
         ops::{Invert, LinearCombination, Reduce},
         subtle::CtOption,
-        AffineArithmetic, AffineXCoordinate, Field, FieldBytes, Group, ProjectiveArithmetic,
-        ProjectivePoint, Scalar, ScalarArithmetic,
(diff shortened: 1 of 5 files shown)


ecdsa/src/hazmat.rs
```diff
@@ -216,8 +217,35 @@ where
...
+            return Err(Error::new());
+        }
+
+        let mut field_bytes = FieldBytes::<Self>::default();
+
+        match prehash.len().cmp(&Self::UInt::BYTE_SIZE) {
+            cmp::Ordering::Equal => field_bytes.copy_from_slice(prehash),
+            cmp::Ordering::Less => {
+                // If prehash is smaller than the field size, pad with zeroes
+                field_bytes[..prehash.len()].copy_from_slice(prehash);
```
sorah, line 239: I'm not confident enough but if I read RFC correctly I believe this should be like:

```rust
field_bytes[(field_bytes.len()-prehash.len())..].copy_from_slice(prehash);
```

https://datatracker.ietf.org/doc/html/rfc6979#section-2.3.2

> otherwise, qlen-blen bits (of value zero) are added to the left of the sequence

the present implementation adds zero on the right so the prehash gets "left shift"'d, while informal description in RFC says:

> in many  software implementations of arithmetics on big integers, that  division is equivalent to a "right shift" by blen-qlen bits.

sorah: Opened a PR for this

➡️  https://github.com/RustCrypto/signatures/pull/547

Issue #520, opened 2022-09-05
DSA: `Verifier` trait not implemented 


Geobert:
Hi,

I'm in a case where I need to verify a signature done on a whole &[u8] rather than a digest and I see that the relevant method is in the `Verifier` trait but it is not  implemented anywhere in the `dsa` crate.

Is that an oversight or it is something just waiting for a contribution? Or am I missing something?

Regards,

aumetra:
If you have a while message aggregated inside a byte slice and you want to verify it, you can do the following (replace `Sha1` with whatever hashing algorithm you use):

```rust
let is_valid = verifying_key.verify_digest(Sha1::new_with_prefix(data)).is_ok();
```

This could maybe be added as to the documentation as reference.

An implementation of the `Verifier` trait would do roughly the same thing but default to the preferred hashing algorithm for the signature scheme. `DigestVerifier` is more flexible and lets you choose the hashing algorithm you want to use.

Geobert:
Fair enough, I have to add dsa support after I've done ed25519 and the latter had `verify` so that confused me. 

Thanks a lot for your answer!

aumetra:
There shouldn't be any complications. The preferred hashing algorithm for DSA is SHA-1, if I'm not mistaken?

lumag:
[RFC 5754](https://www.rfc-editor.org/rfc/rfc5754) defines DSA with SHA-224 and SHA256.

cobratbq:
I had a quick look at this one. It seems RFC 5754 assumes an identifier is included to indicate what hash algorithm was used. So for a plain signature as used for `Signer` and `Verifier` this would be an issue. If DSA is used with SHA-1 by default, but then this is disallowed, considered unsafe for many cases. So any choice for the implementation of `Signer` and `Verifier` is rather arbitrary IIUC. (And, if the default changes at some later point, implementations that upgrade will suddenly find themselves operating with different parameters, which may be a source of problems in itself, especially with interop.)

lumag:
I'd follow what was done for the RSA crate: turn the `Digest` into the generic type argument for the `Signer` and `Verifier` implementations:

```Rust
pub struct Signature {...}
impl signature::Signature for Signature {....}
pub struct SigningKey<D> where D: Digest {...}
impl<D> RandomizedSigner<Signature> for SigningKey<D> where D: Digest {....}
#[cfg(feature = "hazmat")]
[...]

PR #558, opened 2022-10-20
dsa: expose signing and verifying of prehashed hash value

cobratbq:
Expose the ability for the user to provide a prehashed hash-value for signing/verification. In rare cases, this is needed in other protocols. I have attempted to solve this with a custom Digest implementation, but eventually ran into (forced) finalization which affects the hash value.

### DIFF
--- a/dsa/Cargo.toml
+++ b/dsa/Cargo.toml
@@ -22,6 +22,7 @@ opaque-debug = "0.3"
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
 rand = { version = "0.8", default-features = false }
 rfc6979 = { version = "0.3", path = "../rfc6979" }
+sha2 = "0.10"
 signature = { version = ">= 1.6.4, < 1.7", default-features = false, features = ["digest-preview", "rand-preview", "hazmat-preview"] }
 zeroize = { version = "1.5", default-features = false }
 
@@ -30,4 +31,3 @@ pkcs8 = { version = "0.9", default-features = false, features = ["pem"] }
 rand = "0.8"
 rand_chacha = "0.3"
 sha1 = "0.10"
-sha2 = "0.10"
--- a/dsa/Cargo.toml
+++ b/dsa/Cargo.toml
@@ -22,7 +22,7 @@ opaque-debug = "0.3"
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
 rand = { version = "0.8", default-features = false }
 rfc6979 = { version = "0.3", path = "../rfc6979" }
-signature = { version = ">= 1.5, < 1.7", default-features = false, features = ["digest-preview", "rand-preview"] }
+signature = { version = ">= 1.6.4, < 1.7", default-features = false, features = ["digest-preview", "rand-preview", "hazmat-preview"] }
 zeroize = { version = "1.5", default-features = false }
 
 [dev-dependencies]
--- a/dsa/Cargo.toml
+++ b/dsa/Cargo.toml
@@ -22,7 +22,7 @@ opaque-debug = "0.3"
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
 rand = { version = "0.8", default-features = false }
 rfc6979 = { version = "0.3", path = "../rfc6979" }
-sha2 = "0.10"
+sha2 = { version = "0.10", default-features = false }
 signature = { version = ">= 1.6.4, < 1.7", default-features = false, features = ["digest-preview", "rand-preview", "hazmat-preview"] }
 zeroize = { version = "1.5", default-features = false }
 
--- a/dsa/src/signing_key.rs
+++ b/dsa/src/signing_key.rs
@@ -86,6 +86,7 @@ impl SigningKey {
 
         let n = (q.bits() / 8) as usize;
         let block_size = hash.len(); // Hash function output size
+        // FIXME shouldn't `hash.len() < n` be a hard error (bad API use)? According to DSA documentation: "An approved hash function, as specified in FIPS 180, shall be used during the generation of key pairs and digital signatures. When used during the generation of an RSA key pair (as specified in this Standard), the length in bits of the hash function output block shall meet or exceed the security strength associated with the bit length of the modulus n (see SP 800-57)."
 
         let z_len = min(n, block_size);
         let z = BigUint::from_bytes_be(&hash[..z_len]);
--- a/dsa/src/signing_key.rs
+++ b/dsa/src/signing_key.rs
@@ -66,6 +66,16 @@ impl SigningKey {
         &self.x
     }
 
+    /// `sign_hash` signs a pre-hashed value.
+    #[must_use]
+    pub fn sign_hash<D>(&self, hash: &[u8]) -> Option<Signature>
+    where
+        D: Digest + BlockSizeUser + FixedOutputReset,
+    {
+        let k_kinv = crate::generate::secret_number_rfc6979::<D>(&self, hash);
+        self.sign_prehashed(k_kinv, hash)
+    }
+
     /// Sign some pre-hashed data
     fn sign_prehashed(&self, (k, inv_k): (BigUint, BigUint), hash: &[u8]) -> Option<Signature> {
         let components = self.verifying_key().components();
--- a/dsa/src/signing_key.rs
+++ b/dsa/src/signing_key.rs
@@ -12,7 +12,7 @@ use pkcs8::{
     AlgorithmIdentifier, DecodePrivateKey, EncodePrivateKey, PrivateKeyInfo, SecretDocument,
 };
 use rand::{CryptoRng, RngCore};
-use signature::{DigestSigner, RandomizedDigestSigner};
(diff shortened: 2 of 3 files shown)


cobratbq:
@aumetra can you confirm that there is no component of randomness in this logic? The `crate::generate::secret_number_rfc6979` does not seem to depend on randomness.

tarcieri:
@aumetra for now you can just do the deterministic API.

RFC6979 supports additional randomness, but that can be added as a followup.

aumetra:
> can you confirm that there is no component of randomness in this logic?
Yes, the logic of the function (as it currently is) is fully deterministic

tarcieri:
@cobratbq looks like one clippy nit for a needless borrow

tarcieri:
> Yes, the logic of the function (as it currently is) is fully deterministic

@aumetra it actually also supports using an RNG:

https://github.com/RustCrypto/signatures/blob/a440df135ad0feaa17978b1aabeee9d4dd55eafa/dsa/src/signing_key.rs#L108-L124

aumetra:
Oh yeah, I was talking about the function `crate::generate::secret_number_rfc6979` explicitly though, since they only asked about that one

cobratbq:
@tarcieri @aumetra quick follow-up, I'd appreciate your feedback

1. I added an assertion to ensure hash length is at least the same strength as modulus. Before it just took the minimum of two numbers.
2. `verify_prehash` is implemented, seems straight-forward enough. No issue there.
3. `sign_prehash`, IIUC, cannot be implemented with this combination of trait `PrehashSigner<Signature>` and the requirement of a `D: Digest + BlockSizeUser + FixedOutputReset` for `crate::generate::secret_number_rfc6979`.

Do you have an idea what to do with __(3)__? The type for `D` has to be somewhere, and on the trait seems to make most sense. (Branch, currently, does not build but does illustrate the choke point.)

tarcieri:
@cobratbq `DigestSigner` is already the trait which has a `D` parameter.

The ECDSA crate uses a curve-specific digest when implementing `PrehashSigner`:

https://github.com/RustCrypto/signatures/blob/3a13e77/ecdsa/src/sign.rs#L268

Failing that you could just use e.g. SHA-256, or some other SHA2-family digest which has test vectors in RFC6979.

cobratbq:
@tarcieri I've been looking into the ecdsa implementation. From what I can tell, I suspect that the `PrehashSigner` trait is currently ill-defined: you rely on that trait to define deterministic signing behavior that needs to have a `Digest` defined, however it is not part of this trait.

It works for `ecdsa` because of luck, i.e. it is coincidentally already defined on the implementation. For `dsa` that won't work, because the spec allows for any number of approved digests, while `PrehashSigner` requests/requires an input parameter `prehash` with a run-time bound: `bit-length at least as large as bit-length of digest`. (Also, `DigestSigner` is not necessarily present if itself not required.)

Is this something you would want to change/fix or should I work around it, like effectively (probably by accident) happened with `ecdsa`?

aumetra:
The logic in `ecdsa` isn't really ill-defined. Incompatible with implementations that allow free definition of the used digest function, sure, but not wrong.
The crux is that ECDSA has recommended hash digests for each curve. P256 -> SHA-256, P384 -> SHA-384, you get the picture.  
To cite the FIPS 186-4:

> It is recommended that the security strength associated with the bit length of n and the security
strength of the hash function be the same unless an agreement has been made between
participating entities to use a stronger hash function

The `ecdsa` crate uses the associated type the recommended digest to generate the ad-hoc key with.

DSA however doesn't really have a "recommended" hash digest. Most of the standards reference SHA-1 which is actually pretty awful because, [you know](https://sha-mbles.github.io/).  
This is actually tangentially related to a discussion in #520, where we are practically still trying to figure out which digest should be used for a potential `Signer`/`Verifier` implementation.

---

Also regarding your change where you added the assertion on the hash length and the size of parameter `q`, if I read the standard correctly, this restriction only applies to RSA (specifically RSA key pair generation).  

The full quote is as follows:

> An approved hash function, as specified in FIPS 180, shall be used during the generation of key
pairs and digital signatures. When used during the generation of an RSA key pair (as specified in
this Standard), the length in bits of the hash function output block shall meet or exceed the
security strength associated with the bit length of the modulus n (see SP 800-57).

The quote can be found in FIPS 186-4, under section 5.1 "RSA Key Pair Generation"

tarcieri:
@cobratbq the current shape of the `ecdsa` crate is based on years of iteration and hard-won knowledge about how RFC6979 is deployed alongside ECDSA in many real-world scenarios. In fact, we have to accommodate cases where the digest used for instantiating HMAC-DRBG is different from the one used to hash the message.

We previously did what I suspect you're suggesting: always use the `D` parameter of `DigestSigner` to instantiate HMAC-DRBG. That felt great at first, but we ended up with many, many requests to match existing test vectors for systems which use two different hashes, namely Keccak256 for message hashing and SHA-256 for RFC6979 as commonly used in the Ethereum ecosystem.

Unless you have such specific requirements as well, you can instantiate RFC6979 with the secure cryptographic digest of your choice, which is why I recommended SHA-256 as a reasonable default. Again, it doesn't have to match what was used to hash the message, and really the only party who can discern what hash was used was the signer.

`PrehashSigner` doesn't have a `D` parameter by design: the `DigestSigner` trait provides that. If you really, really want to have a generic parameter for `D` when using `PrehashSigner`, it needs to be defined on the type, not the trait.

tarcieri:
> I would define a `D` type on `PrehashSigner`.

Again, `PrehashSigner` doesn't have a `D` parameter *by design*. `PrehashSigner` is explicitly designed to be able to handle different hash inputs at *runtime*. The `ecdsa` crate implements this logic:

See:

- https://github.com/RustCrypto/signatures/blob/8c73683/ecdsa/src/sign.rs#L264
- https://github.com/RustCrypto/signatures/blob/8c73683/ecdsa/src/hazmat.rs#L227-L251

Even if we were to abandon that requirement, a `D` parameter doesn't make sense in all of the potential contexts where `PrehashSigner` would be used. And again, there's already a similar trait with a `D` parameter: `DigestSigner`.

That's why I suggested that if you really, really want to expose it as a user-configurable parameter, it needs to be on the *type*, not the trait.

cobratbq:
@tarcieri just to check, I propose a `D` to pass on to `crate::generate::secret_number_rfc6979::<D>(&self, prehash)`. The value `prehash` is a slice of unknown length and that is fine, i.e. works exactly as expected.

tarcieri:
@cobratbq aah, that's fine

tarcieri:
> regarding the HMAC-DRBG, would it be better to: (1.) hardcode for DSA? Or (2.) define a generic D that is specific to the use-cases of variable or multiple digests (as discussed above)?

If #520 is resolved, you can use the same approach as ECDSA, leveraging `dsa::Signature::Digest`.

> And are there preferences regarding: modifying PrehashSigner vs. introducing an (intermediary) trait such that the generic type can be defined there? (given interoperability, backwards-compatibility, etc.)

If you really feel strongly about this issue, please open an issue on https://github.com/RustCrypto/traits/issues with your suggestion.

If you're suggesting adding an additional trait, please note there's something of a "trait explosion" to handle all possible cases: `PrehashSigner` has an accompanying `RandomizedPrehashSigner` that is parameterized by an RNG, `DigestSigner` also has a `RandomizedDigestSigner`, and all of these need accompanying traits in [`async-signature`](https://docs.rs/async-signature).

Note that there are other things we'd potentially like to add, like domain separation, which would further compound this trait explosion.

So I think the bar would be pretty high to add an additional trait, especially given `DigestSigner` and `PrehashSigner` already exist.

cobratbq:
I get the impression you have some reservations. I understand, especially given another issue is involved.

My goal is to have prehash signing and verification exposed as part of the API. Given the complications as discussed above, what do you recommend?

tarcieri:
Use `Sha256` for now and we can circle back on a more generalized solution in #520

cobratbq:
@aumetra @tarcieri I made the necessary changes. I think my changes have not impacted the other build failures. Can you let me know if there is anything left to do? Also, consider squashing the merge, because the history contains mostly useless confusing changes.

aumetra:
Yep, the build failures seem unrelated. I will have a look

dsa/Cargo.toml
```diff
@@ -22,12 +22,12 @@ opaque-debug = "0.3"
 pkcs8 = { version = "0.9", default-features = false, features = ["alloc"] }
 rand = { version = "0.8", default-features = false }
 rfc6979 = { version = "0.3", path = "../rfc6979" }
-signature = { version = ">= 1.5, < 1.7", default-features = false, features = ["digest-preview", "rand-preview"] }
+sha2 = "0.10"
```
tarcieri, line 25: This is the cause of the build failure. It enables `std` by default: https://github.com/RustCrypto/hashes/blob/1731ced/sha2/Cargo.toml#L30

It succeeded before because the check is `cargo build` which ignores `dev-dependencies`.

You need to do:

```suggestion
sha2 = { version = "0.10", default-features = false }
```

cobratbq: I applied this suggestion, though first locally.

tarcieri (approved):

PR #562, opened 2022-10-30
Bump `signature` to 2.0.0-pre

tarcieri:
Implements the proposed breaking changes to the `signature` crate from https://github.com/RustCrypto/traits/pull/1141

Most notably the `Signature` trait has been replaced with a `SignatureEncoding` trait which permits an internally structured signature representation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,7 +142,7 @@ dependencies = [
 
 [[package]]
 name = "dsa"
-version = "0.4.2"
+version = "0.5.0-pre"
 dependencies = [
  "digest 0.10.5",
  "num-bigint-dig",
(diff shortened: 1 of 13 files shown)


tarcieri:
cc @aumetra @cobratbq @lumag

tarcieri:
also cc @rozbb for the `ed25519` crate changes

lumag:
Ok. For me the `s.to_der()` still doesn't fit the generic code. What I'd probably like to see is the `Signer` / `Verifier` implementations for the `p384::der::Signature` struct.

lumag:
If only I could get rid of the `F: Fn(&Signature) -> Vec<u8>` closure in my code :-)

tarcieri:
@lumag you should be able to use `SignatureEncoding::to_vec` now

lumag:
@tarcieri no, because of `signature.to_der()` for ECDSA. X.509 cerfificates use the DER encoding.

tarcieri:
A forthcoming PR to let you sign directly with `ecdsa::der::Signature` should take care of that?

lumag:
Hopefully

PR #566, opened 2022-11-01
ecdsa: Implementation of recoverable signatures

### LINKED ISSUE #525: ecdsa: `VerifyingKey` recovery support
We could potentially adapt the [ECDSA public key recovery algorithm implemented in `k256`](https://github.com/RustCrypto/elliptic-curves/blob/c48ae66/k256/src/ecdsa/recoverable.rs#L177-L193) into a generic implementation which works with any elliptic curve.

It could use [`ecdsa::RecoveryId`](https://docs.rs/ecdsa/latest/ecdsa/struct.RecoveryId.html) as an input parameter, leaving exactly how to encode that as an exercise to the user.

I would probably suggest implementing the algorithm as something like `VerifyingKey::recover(msg, sig, rec_id)`.

aumetra:
Adds an initial implementation of Ethereum-style recoverable signatures that's generic over the underlying curve.  
This is mostly a port of the code in the `k256` crate adapted to work generically.  

"Marked ready for review" checklist:

- [x] Implementation of the algorithm to recover verifying keys
- [x] Implementation of `SignatureEncoding` for the recoverable signature
- [ ] More granular `cfg` flags 
- [ ] Implementation of the signer and verifier traits on the key types for the new `RecoverySignature` 
- [ ] Update the internal ECDSA logic to actually emit a `RecoveryId` (instead of always returning `None`)

Trial recovery was tested with P256 and P384 and their respective preferred digest  

Closes #525

### DIFF
--- a/ecdsa/src/lib.rs
+++ b/ecdsa/src/lib.rs
@@ -77,7 +77,7 @@ mod sign;
 #[cfg(feature = "verify")]
 mod verify;
 
-pub use crate::recovery::RecoveryId;
+pub use crate::recovery::*;
 
 // Re-export the `elliptic-curve` crate (and select types)
 pub use elliptic_curve::{self, sec1::EncodedPoint, PrimeCurve};
--- a/ecdsa/src/recovery.rs
+++ b/ecdsa/src/recovery.rs
@@ -69,21 +69,19 @@ impl From<RecoveryId> for u8 {
     }
 }
 
-#[cfg(all(feature = "arithmetic", feature = "verify"))]
+#[cfg(feature = "verify")]
 mod signature {
-    use core::ops::Add;
-
     use crate::{
         hazmat::{DigestPrimitive, VerifyPrimitive},
         RecoveryId, SignatureSize, VerifyingKey,
     };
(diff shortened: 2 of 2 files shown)


tarcieri:
@aumetra I was just about to take a look at implementing something like this

I'd prefer not to define a separate `Signature` type, but adding the relevant methods to `Signature` and `VerifyingKey` itself

aumetra:
Sure, the implementation shouldn't be too difficult to move to the existing `Signature` type. How would we handle serialisation?  

Should the `SignatureEncoding` implementation just emit the recovery ID byte or should it only be included via a special associated method like a theoretical `to_bytes_recoverable`?

tarcieri:
@aumetra that was my plan, yes.

I was planning on working on it this weekend, as part of a larger effort to move to `signature` 2.0

aumetra:
Oh I see, if you feel more comfortable doing it yourself you can just take whatever I implemented so far (if it's of any use to you) and base it off of that, I don't mind

tarcieri:
Closing in favor of #576

PR #593, opened 2022-12-02
ed25519: change Signature::from_bytes to take sized array

mina86:
Change Signature::from_bytes to take `[u8; 64]` as argument rather
than `&[u8]` which:
- provides consistency with KeypairBytes::from_bytes,
- encodes the length in type system allowing clients to assert length
  at compile time and
- removes run-time length check.

At the same time, since there’s already TryFrom<&[u8]> implementation
for Signature, no features are lost and clients who want to create the
signature from a slice can do it via Signature::try_from.

This does change the API but since the crate is currently in 2.0
preview stage, that should be fine since API breaking changes are
allowed by symantic versioning at major version upgrade.

### DIFF
--- a/ed25519/src/lib.rs
+++ b/ed25519/src/lib.rs
@@ -292,9 +292,7 @@ impl Signature {
     pub const BYTE_SIZE: usize = 64;
 
     /// Parse an Ed25519 signature from a byte slice.
-    pub fn from_bytes(bytes: &[u8]) -> signature::Result<Self> {
-        let result = bytes.try_into().map(Self).map_err(|_| Error::new())?;
-
+    pub fn from_bytes(bytes: &[u8; Self::BYTE_SIZE]) -> signature::Result<Self> {
         // Perform a partial reduction check on the signature's `s` scalar.
         // When properly reduced, at least the three highest bits of the scalar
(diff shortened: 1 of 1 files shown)


mina86:
see also: https://github.com/dalek-cryptography/ed25519-dalek/pull/230#issuecomment-1334569131
cc: @tarcieri

ed25519/src/lib.rs
```diff
@@ -292,9 +292,7 @@ impl Signature {
     pub const BYTE_SIZE: usize = 64;
 
     /// Parse an Ed25519 signature from a byte slice.
-    pub fn from_bytes(bytes: &[u8]) -> signature::Result<Self> {
-        let result = bytes.try_into().map(Self).map_err(|_| Error::new())?;
-
+    pub fn from_bytes(bytes: [u8; Self::BYTE_SIZE]) -> signature::Result<Self> {
```
tarcieri, line 295: Not taking a reference will use up more stack space making a copy where it could just be a pointer, unless LLVM inlines it.

Elsewhere we generally take a reference for this reason.

tarcieri (approved):

tarcieri:
Thank you!

PR #623, opened 2023-01-21
ed25519: infallible signature parsing

tarcieri:
We've had a couple request to make signature parsing infallible. Presently it checks that the `s` scalar component of the signature has its three highest bits clear, which will reject some but not all signatures where `s` overflows the curve's order.

However, this check will not reject *all* such invalid signatures (at least as presently implemented), and the same check will be performed at the time the signature is verified (which *will* reject all such invalid signatures).

Though we already shipped `ed25519` v2.0.0 and this is a breaking change, that release has been yanked so we can get this change in last minute.

### DIFF
--- a/ed25519/src/lib.rs
+++ b/ed25519/src/lib.rs
@@ -289,6 +289,13 @@ pub type ComponentBytes = [u8; COMPONENT_SIZE];
 pub type SignatureBytes = [u8; Signature::BYTE_SIZE];
 
 /// Ed25519 signature.
+///
+/// This type represents a container for the byte serialization of an Ed25519
+/// signature, and does not necessarily represent well-formed elements of the
+/// respective elliptic curve fields.
+///
+/// Signature verification libraries are expected to reject invalid field
+/// elements at the time a signature is verified.
 #[derive(Copy, Clone, Eq, PartialEq)]
 #[repr(C)]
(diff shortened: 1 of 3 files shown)


rozbb:
LGTM. Maybe a doc comment somewhere  (either on the struct or on all the methods) that the contents is a dumb byte representation and there is no guarantee it represents a well-formed signature (ie scalar might exceed modulus). And, if exists, a pointer to a function that _will_ let you know if there's an issue with the sig

tarcieri:
Added a comment about the signature being a dumb byte representation in a5512a0.

> And, if exists, a pointer to a function that will let you know if there's an issue with the sig

In the comment I noted signature verifiers are expected to perform these checks.

This commit removes the only logic this crate previously had for performing such checks, which as noted was incomplete at best.

rozbb:
> and does not necessarily represent well-formed elements of the respective elliptic curve fields.

Better: and does not necessarily represent well-formed field or curve elements

It's still not entirely true because any byte sequence is a valid compressed Edwards Y representation. A more pendantic thing would be:

`s` does not necessarily represent a well-formed field element

tarcieri:
@rozbb updated the wording in #624

rozbb:
Looks good!

PR #627, opened 2023-01-21
ed25519: bump `ring-compat` to v0.6

tarcieri:
This release bumps `signature` to v2


tarcieri:
Ugh, guess this won't be possible until we bump version to 1.60

PR #679, opened 2023-03-10
Rename additional_data to personalization_string

bluegate010:
Section 10.1.2.3 of [SP 800-90A](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90Ar1.pdf) describes the instantiation function and shows that this argument should be called `personalization_string`. `additional_data` is something provided during bit generation, so if used it should be provided to the `fill_bytes` function of this library.

### DIFF
--- a/rfc6979/src/lib.rs
+++ b/rfc6979/src/lib.rs
@@ -107,7 +107,7 @@ where
     D: Digest + BlockSizeUser + FixedOutputReset,
 {
     /// Initialize `HMAC_DRBG`
-    pub fn new(entropy_input: &[u8], nonce: &[u8], additional_data: &[u8]) -> Self {
+    pub fn new(entropy_input: &[u8], nonce: &[u8], personalization_string: &[u8]) -> Self {
         let mut k = SimpleHmac::new(&Default::default());
         let mut v = GenericArray::default();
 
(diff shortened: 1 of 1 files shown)


rfc6979/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub type ByteArray<Size> = GenericArray<u8, Size>;
 /// - `x`: secret key
 /// - `n`: field modulus
 /// - `h`: hash/digest of input message: must be reduced modulo `n` in advance
-/// - `data`: additional associated data, e.g. CSRNG output used as added entropy
+/// - `data`: personalization string, e.g. CSRNG output used as added entropy
```
tarcieri, line 63: Though this name is fine for the parameter to `HmacDrbg::new`, RFC6979 does not refer to this parameter as "personalization string".

It's defined in [RFC6979 § 3.6](https://www.rfc-editor.org/rfc/rfc6979#section-3.6):

```
   o  Additional data may be added to the input of HMAC, concatenated
      after bits2octets(H(m)):

         K = HMAC_K(V || 0x00 || int2octets(x) || bits2octets(h1) || k')
```

RFC6979 uses the terminology "additional data" and `k'` to refer to it. That it's passed as the HMAC-DRBG personalization string is an implementation detail users of this API are unlikely to be aware of.

bluegate010: Ack, thanks, removed this line from the PR.

tarcieri:
Cool, looks good now, thanks!

Issue #584, opened 2022-11-06
ecdsa: `VerifyingKey::recover_*` should handle `R.x` overflowing curve's order


tarcieri:
The `r` scalar component of an ECDSA signature is computed by lifting the affine x-coordinate of the curve point 𝑹 = 𝑘×𝑮 into an integer and reducing it modulo the curve's order into an element of the scalar field.

Though occurring infrequently (sometimes with a vanishingly small probability), this condition results in `RecoveryId` values of 2 or 3.

Presently only `RecoveryId` values of 0 or 1 are supported, and the others will result in an error when recovering the verifying key.

tarcieri:
Fixed in #680

PR #680, opened 2023-03-17
Handle the reduced x case in public key recovery

jonas-lj:
This PR handles the case where the x-coordinate of r is reduced mod N in both recoverable signing and public key recovery. The chance that the x is reduced during signing with a randomly chosen k is negligible, so the changes to the signing method could be omitted.

Closing #584

### DIFF
--- a/ecdsa/src/recovery.rs
+++ b/ecdsa/src/recovery.rs
@@ -285,7 +285,7 @@ where
         let (mut r, s) = signature.split_scalars();
         let z = <Scalar<C> as Reduce<C::Uint>>::reduce_bytes(&bits2field::<C>(prehash)?);
         if recovery_id.is_x_reduced() {
-            r = r + &CurveArithmetic::ORDER;
+            r = r + &C::ORDER;
         };
         let R = AffinePoint::<C>::decompress(&r.to_repr(), u8::from(recovery_id.is_y_odd()).into());
 
--- a/ecdsa/src/recovery.rs
+++ b/ecdsa/src/recovery.rs
@@ -13,10 +13,11 @@ use {
 use {
     crate::{hazmat::VerifyPrimitive, VerifyingKey},
     elliptic_curve::{
+        bigint::CheckedAdd,
         ops::{LinearCombination, Reduce},
         point::DecompressPoint,
         sec1::{self, FromEncodedPoint, ToEncodedPoint},
-        AffinePoint, FieldBytesSize, Group, PrimeField, ProjectivePoint,
+        AffinePoint, FieldBytesEncoding, FieldBytesSize, Group, PrimeField, ProjectivePoint,
     },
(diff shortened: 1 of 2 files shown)


ecdsa/src/recovery.rs
```diff
@@ -285,7 +285,17 @@ where
...
-        let R = AffinePoint::<C>::decompress(&r.to_repr(), u8::from(recovery_id.is_y_odd()).into());
+
+        let mut r_bytes = r.to_repr();
+        if recovery_id.is_x_reduced() {
+            let restored = C::Uint::decode_field_bytes(&r_bytes).checked_add(&C::ORDER);
+            // No reduction should happen here if r was reduced
+            if restored.is_none().into() {
+                return Err(Error::new());
+            }
+            r_bytes = restored.unwrap().encode_field_bytes();
```
tarcieri, line 296: You can avoid the `unwrap` with something like:

```suggestion
            match C::Uint::decode_field_bytes(&r_bytes).checked_add(&C::ORDER).into() {
                Some(restored) => {
                    r_bytes = restored.unwrap().encode_field_bytes();
                }
                None => {
                    // No reduction should happen here if r was reduced
                    return Err(Error::new());
                }
             }
```

tarcieri:
Awesome, thanks!

ecdsa/src/recovery.rs
jonas-lj: That's a good idea! Have done something along those lines in https://github.com/RustCrypto/signatures/pull/680/commits/2730286b91d3fb6b70c1ec235d7c5294c049f96c.

ecdsa/src/recovery.rs
```diff
@@ -285,7 +285,18 @@ where
     ) -> Result<Self> {
         let (r, s) = signature.split_scalars();
         let z = <Scalar<C> as Reduce<C::Uint>>::reduce_bytes(&bits2field::<C>(prehash)?);
-        let R = AffinePoint::<C>::decompress(&r.to_repr(), u8::from(recovery_id.is_y_odd()).into());
+
+        let mut r_bytes = r.to_repr();
+        if recovery_id.is_x_reduced() {
+            r_bytes = match Option::<C::Uint>::from(
```
tarcieri, line 291: Is there some reason you can't assign `r_bytes` in the `Some` match arm? It looks like that's the only case that matters.

I guess it couldn't infer `Option` with `into()`? If not, I would think `Option::from` would still work without the explicit `C::Uint` annotation for the generic type parameter.

jonas-lj: It is possible to assign inside the some/match arm. I'll fix that.

I thought it could infer the type too, but the compiler complains that it can't infer the type of `restored` without the explicit generic type parameter.

tarcieri (approved):

PR #686, opened 2023-04-02
ecdsa: add RFC5758 OID support

tarcieri:
Adds support for obtaining the RFC5758 OIDs associated with the digest algorithm used to compute a signature by introducing a new `SignatureWithOid` type.

The OID for a particular signature is vicariously available via the newly introduced `spki::DynAssociatedAlgorithmIdentifier` trait, which provides a generic interface for applications like X.509 certificate building.

The OID can either be specified explicitly or inferred from a `Digest` type which also impls `AssociatedOid`.

cc @baloo @lumag

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,6 +37,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b645a089122eccb6111b4f81cbc1a49f5900ac4666bb93ac027feaecf15607bf"
 
 [[package]]
+name = "base64ct"
+version = "1.6.0"
+source = "git+https://github.com/RustCrypto/formats.git#fd069a6ff7a9f5abfd3d797b51f50ca0a5307e44"
+
+[[package]]
 name = "bincode"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


lumag:
This approach will not work for the RSA-PSS. We need a `signing_key.algorithm_identifier()`, because any other object (including signature) doesn't have a knowledge of `salt_len`. So, having this on `Signature<C>` type is syntactically correct, but kind of defeats the purpose for me.

tarcieri:
Hmm, we may need a more specific trait in the `signature` crate for that then.

There are two relevant algorithm identifiers here: the one that identifies the public key cryptosystem, and the one that identifies the signature algorithm.

If we use `AssociatedAlgorithmIdentifier` for the latter, we have no way to query the former.

lumag:
Agree. Would it be fine to have it in `spki` where all other traits are defined? Or would you prefer to have that in the `signature` crate? Anyway, this will create a dependency from one crate to another (which we don't have ATM).
cc @baloo

tarcieri:
`spki` is probably a better idea actually, good suggestion

tarcieri:
Perhaps `SignatureAlgorithmIdentifier` for a trait name?

lumag:
This is what I had in mind.

tarcieri:
Going to go ahead and merge this but we can continue to iterate (on e.g. `SignatureAlgorithmIdentifier`) before the next release

PR #687, opened 2023-04-02
ecdsa: implement AssociatedAlgorithmIdentifier support

lumag:
Implement AssociatedAlgorithmIdentifier support for the SigningKey if the curve supports this trait.

### DIFF
--- a/ecdsa/Cargo.toml
+++ b/ecdsa/Cargo.toml
@@ -34,6 +34,7 @@ default = ["digest"]
 alloc = ["elliptic-curve/alloc", "signature/alloc"]
 std = ["alloc", "elliptic-curve/std", "signature/std"]
 
+algorithm-identifier = ["elliptic-curve/pkcs8"]
 arithmetic = ["elliptic-curve/arithmetic"]
 dev = ["arithmetic", "digest", "elliptic-curve/dev", "hazmat"]
 digest = ["signature/digest"]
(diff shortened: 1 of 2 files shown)


ecdsa/Cargo.toml
```diff
@@ -34,6 +34,7 @@ default = ["digest"]
 alloc = ["elliptic-curve/alloc", "signature/alloc"]
 std = ["alloc", "elliptic-curve/std", "signature/std"]
 
+algorithm-identifier = ["elliptic-curve/pkcs8"]
```
tarcieri, line 37: Since it needs `pkcs8` anyway, why not just gate this on `pkcs8`?

lumag: I thought that is not that logical, as one is not directly related to another, but why not?
I used `pkcs8` as a way to get `spki` without adding direct dependency.

ecdsa/src/signing.rs
```diff
@@ -539,3 +542,13 @@ where
...
 }
+
+#[cfg(feature = "algorithm-identifier")]
+impl<C> AssociatedAlgorithmIdentifier for SigningKey<C>
+where
+    C: PrimeCurve + AssociatedAlgorithmIdentifier + CurveArithmetic,
+    Scalar<C>: SignPrimitive<C>,
+    SignatureSize<C>: ArrayLength<u8>,
+{
+    const ALGORITHM_IDENTIFIER: AlgorithmIdentifierRef<'static> = C::ALGORITHM_IDENTIFIER;
```
tarcieri, line 553: I think instead of this we can add a generic impl of `AssociatedAlgorithmIdentifier` to `SecretKey<C>` and `PublicKey<C>` in the `elliptic-curve` crate, then propagate that to `SigningKey<C>` and `VerifyingKey<C>`.

PR #689, opened 2023-04-03
ecdsa: add `SignatureWithOid`

tarcieri:
Adds a signature type which is paramaterized by an RFC5758-defined OID identifying the variant of ECDSA used to produce the signature.

cc @baloo @lumag

### DIFF
--- a/ecdsa/src/lib.rs
+++ b/ecdsa/src/lib.rs
@@ -103,18 +103,21 @@ use {
 };
 
 #[cfg(feature = "digest")]
-use digest::const_oid::ObjectIdentifier;
+use digest::{
+    const_oid::{AssociatedOid, ObjectIdentifier},
+    Digest,
+};
+
+#[cfg(feature = "pkcs8")]
+use elliptic_curve::pkcs8::spki::{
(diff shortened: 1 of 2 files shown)


tarcieri:
The real benefit of this will be dynamically dispatching the hash function to use for verification, but I've left that for a followup PR.

ecdsa/src/lib.rs
```diff
@@ -500,3 +500,194 @@ where
...
+        AlgorithmIdentifierOwned {
+            oid: self.oid,
+            parameters: None,
+        }
+    }
+}
+
+/// Get the ECDSA OID for a given digest OID.
+#[cfg(feature = "digest")]
+const fn ecdsa_oid_for_digest(digest_oid: ObjectIdentifier) -> Option<ObjectIdentifier> {
```
baloo, line 685: :heart:

tarcieri: It's private, but accessible via constants (with compile-time failures if given a non-SHA-2 OID, at that)

baloo: I was looking how to do that over in RSA, but this is very cool!

PR #695, opened 2023-04-04
ecdsa: follow the DynAssociatedAlgorithmIdentifier changes

lumag:
Return Result from algorithm_identifier().

Depends on https://github.com/RustCrypto/formats/pull/974

### DIFF
--- a/ecdsa/src/lib.rs
+++ b/ecdsa/src/lib.rs
@@ -117,7 +117,9 @@ use elliptic_curve::pkcs8::spki::{
 use serdect::serde::{de, ser, Deserialize, Serialize};
 
 #[cfg(all(feature = "alloc", feature = "pkcs8"))]
-use elliptic_curve::pkcs8::spki::{AlgorithmIdentifierOwned, DynAssociatedAlgorithmIdentifier};
+use elliptic_curve::pkcs8::spki::{
+    self, AlgorithmIdentifierOwned, DynAssociatedAlgorithmIdentifier,
+};
 
 /// OID for ECDSA with SHA-224 digests.
 ///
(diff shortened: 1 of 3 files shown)


lumag:
I can not reproduce the build failure locally. Probably a git revision got cached somewhere and is not updated.

tarcieri:
You need to run `cargo update -p spki`. I would not have expected the build to work locally without doing this.

tarcieri:
Is it possible you're not checking in `Cargo.lock`?

You didn't update it in this PR, so I would expect it to fail because you're using an out-of-date version

ecdsa/src/lib.rs
```diff
@@ -672,11 +674,11 @@ impl<C> DynAssociatedAlgorithmIdentifier for SignatureWithOid<C>
 where
     C: PrimeCurve,
 {
-    fn algorithm_identifier(&self) -> AlgorithmIdentifierOwned {
-        AlgorithmIdentifierOwned {
+    fn algorithm_identifier(&self) -> SpkiResult<AlgorithmIdentifierOwned> {
```
tarcieri, line 677:
```suggestion
    fn algorithm_identifier(&self) -> spki::Result<AlgorithmIdentifierOwned> {
```

lumag: @tarcieri there is no direct dependency on `spki`. Would you prefer:
- add spki as a dep?
- use whatever..spki to make `spki` available as a name
- use full path including elliptic-curves::pkcs8:...

ecdsa/src/lib.rs
```diff
@@ -117,7 +117,9 @@ use elliptic_curve::pkcs8::spki::{
 use serdect::serde::{de, ser, Deserialize, Serialize};
 
 #[cfg(all(feature = "alloc", feature = "pkcs8"))]
-use elliptic_curve::pkcs8::spki::{AlgorithmIdentifierOwned, DynAssociatedAlgorithmIdentifier};
+use elliptic_curve::pkcs8::spki::{
+    AlgorithmIdentifierOwned, DynAssociatedAlgorithmIdentifier, Result as SpkiResult,
```
tarcieri, line 121:
```suggestion
    self, AlgorithmIdentifierOwned, DynAssociatedAlgorithmIdentifier,
```

ecdsa/src/lib.rs
tarcieri: See: https://github.com/RustCrypto/signatures/pull/695/files#r1157862631

ecdsa/src/lib.rs
lumag: ack,

lumag: done

tarcieri (approved):

PR #696, opened 2023-04-04
ecdsa: follow the DynAssociatedAlgorithmIdentifier changes 

baloo:
Fix dependencies of #695

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,4 +12,5 @@ opt-level = 2
 
 [patch.crates-io]
 der = { git = "https://github.com/RustCrypto/formats.git" }
+ecdsa = { path = "./ecdsa" }
 spki = { git = "https://github.com/RustCrypto/formats.git" }
--- a/dsa/Cargo.toml
+++ b/dsa/Cargo.toml
@@ -21,7 +21,7 @@ num-traits = { version = "0.2", default-features = false }
 pkcs8 = { version = "0.10", default-features = false, features = ["alloc"] }
 rfc6979 = { version = "0.4", path = "../rfc6979" }
 sha2 = { version = "0.10", default-features = false }
-signature = { version = "2.0, <2.1", default-features = false, features = ["alloc", "digest", "rand_core"] }
+signature = { version = "2.0, <2.2", default-features = false, features = ["alloc", "digest", "rand_core"] }
 zeroize = { version = "1.6", default-features = false }
 
 [dev-dependencies]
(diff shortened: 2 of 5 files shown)


dsa/Cargo.toml
```diff
@@ -21,7 +21,7 @@ num-traits = { version = "0.2", default-features = false }
 pkcs8 = { version = "0.10", default-features = false, features = ["alloc"] }
 rfc6979 = { version = "0.4", path = "../rfc6979" }
 sha2 = { version = "0.10", default-features = false }
-signature = { version = "2.0, <2.1", default-features = false, features = ["alloc", "digest", "rand_core"] }
+signature = { version = "2.1", default-features = false, features = ["alloc", "digest", "rand_core"] }
```
tarcieri, line 24: `signature` was deliberately locked to `<2.1` because `digest` and `rand_core` updates are gated on minor versions.

Since 2.1 is out and didn't bump `digest` and `rand_core`, it should be changed to:

```toml
signature = { version = "2.0, <2.2", default-features = false, features = ["alloc", "digest", "rand_core"] }
```

tarcieri: (also this all seems unrelated to this PR, and would probably best be done in a separate PR)

baloo: but we get build errors if we don't bump dsa:
```
$ cargo check --all-features
    Updating crates.io index
error: failed to select a version for `signature`.
    ... required by package `ecdsa v0.16.2 (/home/arthur_gautier/work/dev/signatures/ecdsa)`
versions that meet the requirements `^2.1` are: 2.1.0

all possible versions conflict with previously selected packages.

  previously selected package `signature v2.0.0`
    ... which satisfies dependency `signature = "^2.0, <2.1"` of package `dsa v0.6.0 (/home/arthur_gautier/work/dev/signatures/dsa)`

failed to select a version for `signature` which could resolve this conflict
```

baloo: we get dependency errors if we don't bump dsa:
```
$ cargo check --all-features
    Updating crates.io index
error: failed to select a version for `signature`.
    ... required by package `ecdsa v0.16.2 (./ecdsa)`
versions that meet the requirements `^2.1` are: 2.1.0

all possible versions conflict with previously selected packages.

  previously selected package `signature v2.0.0`
    ... which satisfies dependency `signature = "^2.0, <2.1"` of package `dsa v0.6.0 (./dsa)`

failed to select a version for `signature` which could resolve this conflict
```

Cargo.toml
```diff
@@ -12,4 +12,5 @@ opt-level = 2
 
 [patch.crates-io]
 der = { git = "https://github.com/RustCrypto/formats.git" }
```
lumag, line 14: @baloo please drop these two patches. We should not have such overrides after a relase

tarcieri:
Closing in favor of #695

Issue #8, opened 2019-03-25
Missing signatures


newpavlov:
- [ ] NIST P-256, NIST P-384

link2xt:
- [ ] DSA

Requested here: https://github.com/rpgp/rpgp/issues/56

QiangHeisenberg:
Is DSA not implemented yet? I can try to add.

tarcieri:
Nobody's working on DSA. You can go for it.

It might be good to come up with a plan for how it would be implemented (e.g. what crate dependencies would be used, for example), and ideally try to align with https://github.com/RustCrypto/RSA on some of those decisions

lumag:
GOST digital signature alsorithm is missing. However additional care should be taken while implementing it. The standard is clueless with respect to endianness of the used numbers. The implementation should follow the established practice, rather than blindly following the standard.

adamvm:
During build automata with rust nightly:

   Compiling ed25519 v1.3.0 (https://github.com/RustCrypto/signatures.git?tag=ed25519/v1.3.0#8282fcaf)
error[E0405]: cannot find trait `Signature` in crate `signature`
   --> /home/ami/.cargo/git/checkouts/signatures-a6df6df1cc32e789/8282fca/ed25519/src/lib.rs:349:17
    |
349 | impl signature::Signature for Signature {
    |                 ^^^^^^^^^ not found in `signature`

For more information about this error, try `rustc --explain E0405`.
error: could not compile `ed25519` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...

tarcieri:
`nightly` is expected to break in ways that have nothing to do with our crates, nor can we do anything to fix them

tarcieri:
A note about ML-DSA: we now have an ML-KEM implementation at https://github.com/RustCrypto/KEMs/tree/master/ml-kem

It would be good for an ML-DSA implementation to share a common core with `ml-kem`.

cothan:
Hi, I'm interested in implementing ML-DSA.

tarcieri:
cc @bifurcation

bifurcation:
Hey @cothan, happy to collaborate on ML-DSA.  Especially if you're willing to take the lead :)  I haven't looked at ML-DSA much, but my impression is that the algebra is similar to ML-KEM, but the constants are different, as are which bits are public/private.  So you'll probably want to change some things, but hopefully the `ml-kem` crate has good framework to adapt.

[...]

PR #727, opened 2023-07-02
ed448: Ed448 Implementation

ashWhiteHat:
Part of https://github.com/RustCrypto/signatures/issues/8

Hi there.
I implemented ed448.

Implementation and documentation are aligned with ed25519 implementation.

## What I did

- hex display and serialization
- pkcs v1 private/public key encoding and decoding
- seder serializers and deserializers

## Issues

I don't know how to convert pkcs v1 to v2 format so I couldn't implement pkcs v2 enc/dec.

### code duplication

Ed448 implementation is almost the same as Ed25519.
There is huge code duplication.
[PublicKeyBytes](https://github.com/ashWhiteHat/signatures/blob/ashWhiteHat/ed448/ed448/src/pkcs8.rs#L191) is only the difference between Ed448 and Ed25519.

In Ed25519, the public key bytes length is 32, and **type [u8; 32]** supports `Default`.
However, in Ed448, the public key bytes length is 57, and **type [u8; 57]** doesn't support `Default`.
I implemented `PublicKeyBytes` as a struct in Ed448.

If we integrate both formats, we would be able to implement them as a macro.

### crate name

[ed448](https://crates.io/crates/ed448) crate was already taken. I would like to know what name we should name.

## Todo

- [ ] support pkcs v2
- [ ] ignore code duplication or not
- [ ] decide crate name

I would appreciate it if you could confirm.
Thank you.


ed448/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "ed448"
```
tarcieri, line 2: We don't own the `ed448` crate

ashWhiteHat: Exactly.
What should I name?
Is `ed448-sig` fine?

ashWhiteHat: Hi @tarcieri 
Is everything okay?

tarcieri: I can't think of a good name here, sorry. Especially since it would be different from `ed25519`

ashWhiteHat: I see.
`ed448_goldilocks` is still available.

tarcieri: Doesn’t look like it: https://crates.io/crates/ed448-goldilocks

ashWhiteHat: I would say that is `ed448-goldilocks` but `ed448_goldilocks` is available.

ashWhiteHat: Ah
It seems it's not available.

ashWhiteHat: hyphen and underbar are not distinguished.

ashWhiteHat: getting eddsa and integrate ed448 and ed25519

ashWhiteHat: in root directly, there are dsa and ecdsa.

ashWhiteHat: I think it makes sense

ashWhiteHat: eddsa seems available miraculously
https://crates.io/crates/eddsa

tarcieri: Good call! Snagged it

ashWhiteHat: Cool!
How can I proceed this PR?
There are two task.

- implement ed448
- integrate ed448 and ed25519 to eddsa

I think it's better to create PR separately because ed448 is enough big changes.

tarcieri: We still need to come up with an Ed448 crate

ashWhiteHat: Copy that.
I will post here once I come up with.

ashWhiteHat: How about edwards448?
https://crates.io/crates/edwards448

ashWhiteHat: You have edwards25519
https://crates.io/crates/edwards25519

ashWhiteHat: In rfc, it's described as edwards448 and edwards25519
https://datatracker.ietf.org/doc/html/rfc8032

tarcieri: I went ahead and snagged `edwards448`.

The problem is those describe the Edwards form of the elliptic curve, not the digital signature algorithm.

ashWhiteHat: I think I can generalize ed448 and ed25519 signature because this library doesn't have curve arithmetic.
How about implementing `signature/eddsa` which supports edwards curve dsa as the same with ecdsa and importing eddsa from `elliptic-curve/edwards25519` and `elliptic-curve/edwards448`?

tarcieri: We don't currently have or plan to maintain implementations of Curve25519 or Ed448-Goldilocks.

`ed25519` is effectively finished and it would be pretty painful to make any additional breaking changes at this point. We're finally to the place where almost everything has been updated to v2.0 which has taken most of 2023.

ashWhiteHat: Copy that.
Let me think.

ashWhiteHat: How about eddsa448 and renaming ed25519 to eddsa25519?
https://crates.io/search?q=eddsa448

tarcieri: Again, it’s taken almost a year to wrap up the latest `ed25519` upgrade and the work is still incomplete.

I’d really rather avoid making any additional breaking changes to `ed25519` for at least another year after everything is fully upgraded and usable, especially just for supporting Ed448 which is relatively obscure.

Also “EdDSA25519” is not the construction’s name, it’s Ed25519.

tarcieri: I guess just call it `ed448-signature` for now. I can't think of a better name.

ashWhiteHat: Thank you for the discussion.
I named it ed448-signature.

tarcieri (approved):

PR #730, opened 2023-07-08
ecdsa: refactor `Signature` constructors and improve docs

tarcieri:
Implements `from_bytes` in terms of `from_scalars`, rather than the other way around.

This places the logic for checking that `r` and `s` are nonzero inside of the `from_scalars` method.

Additionally improves the documentation to note the various failure cases, i.e. if `r` and/or `s` is out of the range `1..n`, where `n` is the scalar modulus.

### DIFF
--- a/ecdsa/src/hazmat.rs
+++ b/ecdsa/src/hazmat.rs
@@ -100,10 +100,7 @@ where
         // Compute 𝒔 as a signature over 𝒓 and 𝒛.
         let s = k_inv * (z + (r * self));
 
-        if s.is_zero().into() {
-            return Err(Error::new());
-        }
-
+        // NOTE: `Signature::from_scalars` checks that both `r` and `s` are non-zero.
         let signature = Signature::from_scalars(r, s)?;
         let recovery_id = RecoveryId::new(R.y_is_odd().into(), x_is_reduced);
         Ok((signature, Some(recovery_id)))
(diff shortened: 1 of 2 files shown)


ecdsa/src/hazmat.rs
```diff
@@ -100,10 +100,7 @@ where
         // Compute 𝒔 as a signature over 𝒓 and 𝒛.
         let s = k_inv * (z + (r * self));
 
-        if s.is_zero().into() {
-            return Err(Error::new());
-        }
-
+        // NOTE: `Signature::from_scalars` checks that both `r` and `s` are non-zero.
```
tarcieri, line 103: @fjarri perhaps you could add a comment like this in https://github.com/RustCrypto/elliptic-curves/pull/907 ?

ecdsa/src/lib.rs
```diff
@@ -236,8 +240,21 @@ where
...
+    ///
+    /// # Returns
+    /// - `Ok(signature)` if the `r` and `s` components are both in the valid
+    ///   range `1..n` when serialized as concatenated big endian integers.
+    /// - `Err(err)` if the `r` and/or `s` component of the signature is
+    ///   out-of-range when interpreted as a big endian integer.
     pub fn from_scalars(r: impl Into<FieldBytes<C>>, s: impl Into<FieldBytes<C>>) -> Result<Self> {
-        Self::try_from(r.into().concat(s.into()).as_slice())
+        let r = ScalarPrimitive::from_slice(&r.into()).map_err(|_| Error::new())?;
+        let s = ScalarPrimitive::from_slice(&s.into()).map_err(|_| Error::new())?;
```
tarcieri, line 251: In the next release it might make sense to change the `r` and `s` parameters to `impl Into<ScalarPrimitive<C>>` to simplif

PR #731, opened 2023-07-17
ecdsa: extract `hazmat::{sign_prehashed, verify_prehashed}`.

tarcieri:
Extracts generic, reusable functions from the `SignPrimitive` and `VerifyPrimitive` traits.

The main motivation for this is to make it possible for the `*Primitive` trait impls to be composed in terms of a generic implementation, in order to add support for low-S normalization in the `k256` crate.

### DIFF
--- a/ecdsa/src/hazmat.rs
+++ b/ecdsa/src/hazmat.rs
@@ -18,7 +18,7 @@ use elliptic_curve::{generic_array::typenum::Unsigned, FieldBytes, PrimeCurve};
 use {
     crate::{RecoveryId, SignatureSize},
     elliptic_curve::{
-        ff::PrimeField,
+        ff::{Field, PrimeField},
         group::{Curve as _, Group},
         ops::{Invert, LinearCombination, MulByGenerator, Reduce},
         point::AffineCoordinates,
@@ -56,7 +56,7 @@ pub trait SignPrimitive<C>:
     + Reduce<C::Uint, Bytes = FieldBytes<C>>
     + Sized
 where
-    C: PrimeCurve + CurveArithmetic + CurveArithmetic<Scalar = Self>,
+    C: PrimeCurve + CurveArithmetic<Scalar = Self>,
     SignatureSize<C>: ArrayLength<u8>,
 {
     /// Try to sign the prehashed message.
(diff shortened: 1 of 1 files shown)


tarcieri:
Note: the intended use of this is to address https://github.com/RustCrypto/elliptic-curves/issues/908.

I think in the next breaking release of `ecdsa` we can just add low-S normalization support to the `ecdsa` crate, and with that drop the `SignPrimitive`/`VerifyPrimitive` traits.

ecdsa/src/hazmat.rs
```diff
@@ -247,6 +205,93 @@ pub fn bits2field<C: PrimeCurve>(bits: &[u8]) -> Result<FieldBytes<C>> {
...
+    // Compute 𝑹 = 𝑘×𝑮
+    let R = ProjectivePoint::<C>::mul_by_generator(k.as_ref()).to_affine();
+
+    // Lift x-coordinate of 𝑹 (element of base field) into a serialized big
+    // integer, then reduce it into an element of the scalar field
+    let r = Scalar::<C>::reduce_bytes(&R.x());
+    let x_is_reduced = r.to_repr() != R.x();
+
+    // Compute 𝒔 as a signature over 𝒓 and 𝒛.
+    let s = k_inv * (z + (r * d));
```
fjarri, line 253: Perhaps we should check for `r=0` or `s=0` here, at the lowest level, so that every high level user is automatically protected?

tarcieri: That check is performed on the [next line of code when `Signature::from_scalars` is called](https://github.com/RustCrypto/signatures/pull/731/files/928c75bab14778182dd7aa037359d6e746c0e0b4#diff-0328eac1fc4a00e7acac377a2d141c54660be93361d7e010b9e7fdb18c5ad43eR256), which checks both [r and s are non-zero](https://github.com/RustCrypto/signatures/blob/master/ecdsa/src/lib.rs#L253-L255).

fjarri: I think since this function is public API it should return valid data, regardless of whether that data will be later packed into a `Signature`. Or perhaps it could be packed into a `Signature` right away, and if some implementations need to do additional processing, they'll just transform that `Signature`.

tarcieri: I'm confused, you want it to return something other than `Signature`?

tarcieri: I'd like to prevent casual users who are poking around `Signature` from accidentally discovering it, which is why it's filed away here instead.

PR #746, opened 2023-10-10
ed25519: enable `pkcs8/std` feature when `std` is enabled

cGuille:
When we enable the `std` feature and the `pkcs8` dependency is enabled (by the `pem` feature), enable the `std` feature of `pkcs8`.

This makes it easier to work with the re-exported `pkcs8::Error` type because without the `std` feature it does not impl the  `std::error::Error` marker trait.


tarcieri (approved):

cGuille:
Sorry, didn't even say hello!

Thanks for the quick merge :)

tarcieri:
Thanks for the contribution!

PR #772, opened 2024-01-12
Fixed rfc reference in comment

bkstein:
There's no Appendix A in RFC 5912. I think, it's a leftover from the previous reference to RFC 5480.


tarcieri:
Yep, looks good

PR #775, opened 2024-01-17
rfc6979: add P-521 + SHA-512 test vectors

tarcieri:
These now work after #781


rfc6979/src/lib.rs
```diff
@@ -154,12 +158,21 @@ where
...
             self.v = self.k.finalize_reset().into_bytes();
             out_chunk.copy_from_slice(&self.v[..out_chunk.len()]);
         }
 
+        let out_remainder = out_chunks.into_remainder();
+        if !out_remainder.is_empty() {
+            self.k.update(&self.v);
+            self.v = self.k.finalize_reset().into_bytes();
+            out_remainder.copy_from_slice(&self.v[..out_remainder.len()]);
+        }
```
tarcieri, line 174: Prior to #773 it was ensured that the output size would always be equal to the digest size, however now we need to handle the remainder.

https://datatracker.ietf.org/doc/html/rfc6979#appendix-A.1 contains a detailed example I need to step through and also add test cases for

tarcieri:
#781 seems to have uncovered the bug

PR #781, opened 2024-01-17
rfc6979: add K-163 test vector; fix nonaligned use

tarcieri:
RFC6979 Appendix A.1. provides a "Detailed Example" which exercises several edge cases in the protocol:

- `bits2int` for an input which is not byte-aligned
- Rejecting inputs which exceed the modulus

This PR adds what was missing from the previous implementation which assumed inputs were always aligned to the size of the digest output: a constant-time right shift by the number of bits by which the modulus is smaller than a byte-aligned value.


tarcieri:
The first HMAC-DRBG output appears correct (albeit truncated to the byte size of the modulus):

```
0x9305a46de7ff8eb107194debd3fd48aa20d5e7656c
```

However, RFC6979 suggests the correct candidate value is this:

```
k1 = 0x4982D236F3FFC758838CA6F5E9FEA455106AF3B2B
```

The problem is that our candidate value is 168-bits, and our modulus is 163-bits. So, we need to add a right shift with the difference:

```
>> hex(0x9305a46de7ff8eb107194debd3fd48aa20d5e7656c >> 5)
0x4982d236f3ffc758838ca6f5e9fea455106af3b2b
```

...so we need our implementation to count the bit length of the modulus first, then perform a right shift if necessary.

Issue #782, opened 2024-01-17
rfc6979: integrate `crypto-bigint`


tarcieri:
Currently `rfc6979` bundles a tiny, slow, byte-oriented bignum library: https://github.com/RustCrypto/signatures/blob/master/rfc6979/src/ct.rs

Several of the interfaces described in RFC6979 describe an "int" type, perhaps most notably the `bits2int` function (which is partially implemented in the `ecdsa` crate in the form of `bits2field`). However all of the APIs in `rfc6979` are byte/octet-oriented, which can make things a bit awkward.

With `crypto-bigint` now supporting a heap-allocated `BoxedUint` and plans to migrate the `dsa` crate to that (#779), as well as having traits for abstracting over stack-vs-heap allocated big integers in `crypto-bigint` like `Integer` and `Monty`, it becomes possible to make `rfc6979` closer to 1:1 with the APIs described in the RFC while being able to support both the `dsa` and `ecdsa` use cases.

tarcieri:
This is a bit tricky because `crypto-bigint` integers are multiples of 64-bits, and `rfc6979` may need to work on smaller integers (e.g. 521-bit for P-521).

tarcieri:
I'm going to close this for now: it would really require new types in `crypto-bigint` which could represent integers which are smaller than multiples of the limb size.

The existing implementation works on the byte level, which isn't as efficient but at least works.

Issue #779, opened 2024-01-17
dsa: migrate from `num-bigint-dig` to `crypto-bigint`


tarcieri:
The `num-bigint`(-dig) crate(s) implement operations in a non-constant-time manner. Notably they include a [`normalize`](https://github.com/rust-num/num-bigint/blob/367481903a56a89eee2bffa5a27f8b9c75bfd40f/src/biguint.rs#L848-L859) function which strips trailing zeros, which is called all over the place.

The `crypto-bigint` crate has been written from the ground-up paying attention to constant-time, and the latest `crypto-bigint` prereleases support a `BoxedUint` which is suitable for use with `dsa`:

https://docs.rs/crypto-bigint/0.6.0-pre.11/crypto_bigint/struct.BoxedUint.html

Additionally, the latest prerelease of the `crypto-primes` crate support generic operation over `Uint` and `BoxedUint`:

https://docs.rs/crypto-primes/0.6.0-pre.0/crypto_primes/fn.generate_prime_with_rng.html

tarcieri:
PR for this open under #784

PR #784, opened 2024-01-17
dsa: migrate to `crypto-bigint`

aumetra:
This is a draft. The code is incredibly ugly.  
Will clean it up as soon as I got all of the tests to pass.


tarcieri:
See also #779 #782

aumetra:
Small question @tarcieri 

```rust
let params = BoxedMontyParams::new_vartime(Odd::new((**components.p()).clone()).unwrap());
let form = BoxedMontyForm::new((*y).clone(), params);

if *y < two() || form.pow(components.q()).to_montgomery() != BoxedUint::one() {
    return Err(signature::Error::new());
}
```

should be equivalent to the following code

```rust
if y < two() || y.modpow(components.q(), components.p()) != BigUint::one() {
    return Err(signature::Error::new());
}
```

Right? Or did I make a mistake with the `MontyParams` and `MontyForms` API?

tarcieri:
@aumetra I think you may have run into a confusing API we should really fix.

Instead of `to_montgomery()` you probably want `retrieve()` or `from_montgomery()`

(Personally I would've named everything `from_canonical` and `to_canonical`)

aumetra:
Ah. Yeah, that was it. I guess when I wrote the code at an airport I overlooked that in the docs

image

tarcieri:
needs an MSRV bump to 1.73 in `.github/workflows/dsa.yml`

dsa/src/signing_key.rs
```diff
@@ -63,31 +68,34 @@ impl SigningKey {
...
         hash: &[u8],
     ) -> signature::Result<Signature> {
         let components = self.verifying_key().components();
         let (p, q, g) = (components.p(), components.q(), components.g());
         let x = self.x();
 
-        let r = g.modpow(&k, p) % q;
+        let params = BoxedMontyParams::new(Odd::new(k).unwrap());
+        let form = BoxedMontyForm::new((**g).clone(), params);
+        let r = NonZero::new(form.pow(p).retrieve() % q).unwrap();
```
tarcieri, line 87: Looks like the test failure is occurring on line #85, ostensibly due to the `unwrap` and an even `k`.

I'm not sure why `k` would be even. Perhaps there's an issue in how you're implementing RFC6979? But either way, it would be good to return `signature::Error` in that case rather than panicking.

You might also be able to leverage the `rfc6979` crate's implementation of the algorithm now as well, rather than just it's implementation of HMAC-DRBG: https://github.com/RustCrypto/signatures/blob/ef35987cd1bca5029a4d1d2c6bf3e31ddea0da38/rfc6979/src/lib.rs#L77-L113

aumetra: Yep. I'm also asking myself how it could even be even. And yes, the `.unwrap()`s are part of the "ugly code" I mentioned at the top of the PR, heh.  
I am intending to clean it all up with proper error handling and everything after getting it to work.

aumetra: It definitely is an issue with my own implementation but only ever since I moved to `crypto-bigint` since it's odd in the `num-bigint-dig` version of the code.  
I'll attempt to move it over to the newly exposed API.

aumetra: Now to adjusting the capacity of the generator `BoxedUint`..

baloo:
If you have any spare time, I think we could use this rewrite very soon as it should ease up the transition to `rand_core` 0.9 a lot.

tarcieri:
Closing in favor of #906

PR #801, opened 2024-01-31
Add lms to RustCrypto/signatures

incertia:
This adds the `lms` directory to RustCrypto/signatures under the crate name `lms-signature`. The authors of the [original RFC](https://datatracker.ietf.org/doc/html/rfc8554#section-5.4) seem quite happy using the phrase so this keeps the naming in the signatures repository consistent. Hopefully I ported the Github workflow without any issues.

There are two notable limitations to our implementation:

1. We rely on `generic_array::GenericArray` from the older `digest-0.10.7` instead of using the `hybrid_array::Array` approach from the upcoming `digest-0.11` ecosystem.
2. We lack a way of compressing Merkle tree data so we just store the entire tree which causes H15, H20, and H25 variants of LMS to not work due to storage limitations.

This PR also does not modify the main README in any way. I have also refrained from committing any changes to Cargo.lock.


.github/workflows/lms.yml
```diff
@@ -0,0 +1,52 @@
...
+    paths:
+      - ".github/workflows/lms.yml"
+      - "lms/**"
+      - "Cargo.*"
+  push:
+    branches: master
+
+defaults:
+  run:
+    working-directory: .
```
tarcieri, line 13: I would expect this to be:

```suggestion
    working-directory: lms
```

lms/.github/workflows/lms-rust.yml
```diff
@@ -0,0 +1,52 @@
+name: lms-rust
```
tarcieri, line 1: Is this left over? It seems redundant with `.github/workflows/lms.yml`

lms/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "lms-signature"
+version = "0.1.0"
+edition = "2021"
```
tarcieri, line 4:
```suggestion
edition = "2021"
license = "Apache-2.0 OR MIT"
repository = "https://github.com/RustCrypto/signatures/tree/master/lms"
readme = "README.md"
rust-version = "1.73"
categories = ["cryptography"]
keywords = ["crypto", "signature"]
```

lms/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+name = "lms"
+path = "lms/lib.rs"
```
tarcieri, line 10: Hmm, this is a lot of directories named `lms`, i.e. `lms/lms/lms`. Perhaps use a standard `src` directory instead?

lms/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0.75"
```
tarcieri, line 14: `anyhow` is typically used for application-level error handling as opposed to libraries

lms/lms/lib.rs
```diff
@@ -0,0 +1,14 @@
+//! LMS in Rust
```
tarcieri, line 1: FWIW, pretty much all of our crates support `no_std`, though I'm happy to address that as a followup

lms/lms/lms/private.rs
```diff
@@ -0,0 +1,354 @@
...
+use std::ops::Add;
+use typenum::{Sum, U28};
+
+/// Opaque struct representing a LMS public key
+///
+/// Note: there is no requirement to map specific LMS algorithms to specific
+/// LM-OTS algorithms so it must be parametrized. With the algorithms provided
+/// by this crate, this is done via
+/// [LmsSha256M32H10](crate::lms::LmsSha256M32H10)<[LmsOtsSha256N32W4](crate::ots::LmsOtsSha256N32W4)>.
+pub struct PrivateKey<Mode: LmsMode> {
```
tarcieri, line 23: As a recommendation for cross-crate consistency, we generally use the names `SigningKey` and `VerifyingKey`, which are named after their roles / what they do, a convention also adopted in `ed25519-dalek`

lms/Cargo.toml
tarcieri: It also looks like it only has one usage here: https://github.com/RustCrypto/signatures/pull/801/files#diff-4a60ec7cc2aeac6ce05a26ae2a5b3d69d23c199b6d4f9a93f2a2bef5ef90ba88R106

lms/lms/lib.rs
incertia: I think our code is `no_std` compatible but it has not been explicitly verified. We can probably address this some time later.

lms/lms/lms/private.rs
incertia: I have updated the names to `SigningKey` and `VerifyingKey`

lms/.gitignore
```diff
@@ -0,0 +1,2 @@
+Cargo.lock
+target/
```
tarcieri, line 2: This whole file can be removed as the toplevel `.gitignore` should address this (and also we check in `Cargo.lock` to have a deterministic build, which is the new recommendation from the Cargo team)

lms/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## 0.1.0 (2024-01-30)
+- Initial release
```
tarcieri, line 8: Leave this blank for now (the date is already invalid) and we can bump it with an actual release:

```suggestion
```

lms/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "lms-signature"
+version = "0.1.0"
```
tarcieri, line 3: We tend to bump version in a release PR/commit, so for now I'd suggest:

```suggestion
version = "0.0.0"
```

lms/src/lms/private.rs
```diff
@@ -0,0 +1,352 @@
...
+
+        let lms_priv = SigningKey::<LmsSha256M32H5<LmsOtsSha256N32W8>>::new_from_seed(id, seed);
+        let lms_pub = lms_priv.public();
+        assert_eq!(lms_pub.k(), expected_k);
+        assert_eq!(lms_pub.id(), &id);
+    }
+
+    #[test]
+    fn test_kat_2() {
+        let expected_signature = [
```
tarcieri, line 235: Can you provide some information on the provenance of these KATs?

It appears to be [RFC8554 Appendix F](https://datatracker.ietf.org/doc/html/rfc8554#appendix-F) but I'm noticing `test_pk_tree_kat1` appears to be "Test Case 2".

lms/src/lms/public.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+        if self.k == tmp {
+            Ok(())
+        } else {
+            Err(Error::new())
+        }
+    }
+}
+
+/// Converts a [VerifyingKey] into its byte representation
```
tarcieri, line 100: To properly linkify in rustdoc, `VerifyingKey` needs to be surrounded with backticks:

```suggestion
/// Converts a [`VerifyingKey`] into its byte representation
```

lms/src/lms/signature.rs
```diff
@@ -0,0 +1,318 @@
+//! Contains the [Signature] type
```
tarcieri, line 1: Similar concern to [this one](https://github.com/RustCrypto/signatures/pull/801/files#r1480048307) regarding backticks (and elsewhere in this file):

```suggestion
//! Contains the [`Signature`] type
```

lms/src/lms/private.rs
incertia: @tjade273

tjade273:
@tarcieri do you have opinions on large in-file KAT buffers like [this](https://github.com/trailofbits/signatures/blob/8e74c6b877949b27583375ee7619ba2d94159d08/lms/src/lms/private.rs#L234)? 

Debating whether we should
1. Leave as-is
2. Break large KATs out to a separate directory
3. Just include a hash of the actual expected result and compare against that

There will be a couple more large KATs eventually once HSS is implemented, and if we want to support the larger merkle tree sizes then they will start getting really big.

tarcieri:
@tjade273 we generally just check them in, but for very large ones omit them from the resulting `.crate`, e.g.:

https://github.com/RustCrypto/MACs/blob/43cc597/cmac/Cargo.toml#L14

tarcieri (approved):

.github/workflows/lms.yml
```diff
@@ -0,0 +1,52 @@
+name: lms-rust
```
tarcieri, line 1: Noticed this after I approved, would probably just suggest this match the crate name:

```suggestion
name: lms-signature
```

You might also consider naming this file the same as the crate name, i.e. `lms-signature.yml`

tarcieri: Then I remembered we omit it with `ed448`. How about just calling it `lms`?

tarcieri:
utACK, let me know if you'd like to make any additional changes or otherwise we can get this merged

incertia:
We can probably get this merged.

tarcieri (approved):

PR #812, opened 2024-04-04
Add SLH-DSA

tjade273:
This PR adds SLH-DSA to the `signatures` crate

The implementation is feature-complete and passing the KATs for all parameter sets. It should be useable with `no-std`, although the stack usage may be too large for most embedded devices.

Planed improvements:
* Provide zero-copy and streaming interfaces in order to reduce stack usage
* Reduce test run-time / factor into CI and release tests


slh-dsa/Cargo.lock
```diff
@@ -0,0 +1,1120 @@
+# This file is automatically @generated by Cargo.
```
tarcieri, line 1: You can remove this file since it's part of the workspace

slh-dsa/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
+.vscode
```
tarcieri, line 2: You can remove this file since it's configured at a workspace-level

tarcieri:
What is `archive.zip`?

tjade273:
> What is `archive.zip`?

Just the git archive I used to move from the repo I was developing in, accidentally checked in. Removed

slh-dsa/src/xmss.rs
```diff
@@ -0,0 +1,260 @@
+use hybrid_array::{Array, ArraySize};
+use typenum::Unsigned;
+
+use crate::wots::WotsSig;
+use crate::{address, wots::WotsParams};
+use crate::{PkSeed, SkSeed};
+use core::fmt::Debug;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct XMSSSig<P: XMSSParams> {
```
tarcieri, line 10: Following the Rust RFC 430 naming conventions: https://rust-lang.github.io/api-guidelines/naming.html

The types/traits contained in this module should be prefixed `Xmss*` rather than `XMSS*`.

```suggestion
pub struct XmssSig<P: XmssParams> {
```

slh-dsa/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "slh-dsa"
+version = "0.0.1"
+edition = "2021"
+
+[dependencies]
+hybrid-array = {version = "0.2.0-rc.8", features = ["extra-sizes"]}
+typenum = {version = "1.17.0", features = ["const-generics"]}
```
tarcieri, line 8: You can access this as `hybrid_array::typenum`, which also requires `typenum` v1.17, and also enables the `const-generics` feature

tarcieri (approved):
Approved, with a caveat that the review is largely for style and consistency as I have only a cursory understanding of the various parts of this algorithm.

slh-dsa/Cargo.toml
tarcieri: I can fix this up after merging.

Issue #843, opened 2024-08-13
slh-dsa: update to FIPS 205 final


tarcieri:
It's out: https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.205.pdf

cc @tjade273

> ### A.1 Changes From FIPS 205 Initial Public Draft
> The differences from Version 3 of the SPHINCS+ specification described in Appendix A were
> included in the draft version of this standard (FIPS 205 ipd) that was posted on August 24, 2023.
> Based on comments that were submitted on FIPS 205 ipd, the SLH-DSA signature generation and
> verification functions were modified to include domain separation cases in which the message
> is signed directly and in which a digest of the message is signed. The changes were made by
> modifying the inputs to the signing and verification functions (see Algorithms 22, 23, 24, and 25).

DiscordJim:
Anyone currently working on this? Would be thrilled to help out.

tarcieri:
@DiscordJim might wait a day to hear from @tjade273, otherwise go for it

DiscordJim:
@tarcieri Sounds good. Ideally I'd like to work on the 204 implementation first. Do you guys have a Discord or something of the like?

tarcieri:
@DiscordJim you can find some discussion about implementing ML-DSA in #8. There's currently an empty `ml-dsa` crate in-tree. Ideally it would share some code with [`ml-kem`](https://github.com/RustCrypto/KEMs/tree/master/ml-kem). You can [chat about it on our Zulip](https://rustcrypto.zulipchat.com/#narrow/stream/260048-signatures)

tjade273:
Nice! I can take care of this tomorrow morning

tarcieri:
Fixed in #844

PR #844, opened 2024-08-14
Implement changes from FIP 205 Initial Public Draft -> FIPS 205 Final

tjade273:
Addresses #843 

- [X] Implement new context string handling for "pure" algorithms
- [X] Change existing KAT to use deterministic `*_internal` functions specified in the new standard
- [ ] Known-answer tests for top-level signatures with context (not sure these exist yet)
- [x] End-to-end tests for signatures with non-empty context
- [ ] Implement new pre-hash variants (using `DigestSigner`)
- [x] Integrate [NIST CVP KATs](https://github.com/usnistgov/ACVP-Server/tree/f38183487eebff2952da0e5a3441371218acfe3f/gen-val/json-files/SLH-DSA-sigGen-FIPS205)


tjade273:
AFAICT there are no published KATs for the full end-to-end signature (with a context string). I'm not too worried about that, it's a very simple spec change.

I'm going to hold off on the pre-hash `DigestSigner` implementation for a separate feature. The requirement to include the hash OID in the context makes integrating arbitrary digests a pain. I don't expect pre-hash to have much demand right now anyway.

tjade273:
@tarcieri This is ready for review. I'll open a new issue for supporting the pre-hash variant, but now at least the normal variant that 99% of users need matches the final spec.

tarcieri:
> The requirement to include the hash OID in the context makes integrating arbitrary digests a pain. 

You can use the [`AssociatedOid` trait](https://docs.rs/const-oid/latest/const_oid/trait.AssociatedOid.html), namely `AssociatedOid + Digest`.

See this for an example: https://docs.rs/rsa/latest/rsa/pkcs1v15/struct.SigningKey.html

tarcieri:
Updated test vectors are available here:

- https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/SLH-DSA-keyGen-FIPS205
- https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/SLH-DSA-sigGen-FIPS205
- https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/SLH-DSA-sigVer-FIPS205

tjade273:
> You can use the [AssociatedOid trait](https://docs.rs/const-oid/latest/const_oid/trait.AssociatedOid.html), namely AssociatedOid + Digest.

Thanks! Saves me from having to implement the consts myself...

I started working on the Oid/DigestSigner feature but have to go through and update various things to the newest pre-release versions to match the pre-release `digest` crate used by `signature`.

tjade273:
> Updated test vectors are available here:

Yep, already incorporated!

https://github.com/RustCrypto/signatures/pull/844/files#diff-bcecd5d27d356a81e18400b7905fdc36196cc3e0de024579b21b5dea7aa808fd

PR #845, opened 2024-08-16
slh-dsa: fix no_std

erdnaxe:
This patch fixes slh-dsa build for non-std targets.

Without this patch, `cargo build --target thumbv7m-none-eabi --no-default-features` fails due to std usage in `crypto-common`.


tarcieri (approved):
Looks good, and also like we need a CI job to test the build

PR #846, opened 2024-08-16
slh-dsa: fix multiple AsBytes definitions

erdnaxe:
For some reasons, I got the following build error while using `slh-dsa` in a no_std project:
```
[...]
error[E0252]: the name `AsBytes` is defined multiple times
  --> /home/user/.cargo/git/checkouts/signatures-f73d5010bc56a463/cfda25f/slh-dsa/src/address.rs:23:5
   |
22 | use zerocopy::AsBytes;
   |     ----------------- previous import of the macro `AsBytes` here
23 | use zerocopy_derive::AsBytes;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^ `AsBytes` reimported here
   |
   = note: `AsBytes` must be defined only once in the macro namespace of this module

For more information about this error, try `rustc --explain E0252`.
error: could not compile `slh-dsa` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

I don't really understand why `cargo test` in `RustCrypto/signatures/slh-dsa` works fine. Here is a proposition for a fix, but maybe there is a cleaner way to address this issue ?

### DIFF
--- a/slh-dsa/src/address.rs
+++ b/slh-dsa/src/address.rs
@@ -20,7 +20,6 @@ use typenum::U22;
 
 use zerocopy::byteorder::big_endian::{U32, U64};
 use zerocopy::AsBytes;
-use zerocopy_derive::AsBytes;
 
 /// `Address` represents a hash address as defined by FIPS-205 section 4.2
 pub trait Address: AsRef<[u8]> {
@@ -40,7 +39,7 @@ pub trait Address: AsRef<[u8]> {
     }
 }
 
-#[derive(Clone, AsBytes)]
+#[derive(Clone, zerocopy_derive::AsBytes)]
 #[repr(C)]
 pub struct WotsHash {
     pub layer_adrs: U32,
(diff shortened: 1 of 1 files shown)


tarcieri:
It seems like the problem here is that `zerocopy` [conditionally re-exports `zerocopy_derive::AsBytes`](https://docs.rs/zerocopy/latest/zerocopy/derive.AsBytes.html) when its `derive` feature is enabled, but `slh-dsa` is not enabling it.

What probably happened @erdnaxe is some other crate in your project enabled the `zerocopy/derive` feature, triggering the re-export.

If you explicitly enable `zerocopy`'s `derive` feature, then it should re-export the `zerocopy::AsBytes` derive macro, making the direct dependency and imports from `zerocopy-derive` unnecessary.

tarcieri:
I think #847 should take care of it

erdnaxe:
Thank you! I confirm that it fixes the build error.

Issue #854, opened 2024-09-04
slh-dsa: no_std broken due to use of concat()


erdnaxe:
When building a project for thumbv7em-none-eabihf target, cargo fails with the following error:
```
   Compiling slh-dsa v0.1.0
error[E0599]: no method named `concat` found for array `[&[u8]; 4]` in the current scope
   --> /home/redacted/.cargo/registry/src/index.crates.io-6f17d22bba15001f/slh-dsa-0.1.0/src/signing_key.rs:146:56
    |
146 |         let ctx_msg = [&[0], &ctx_len_bytes, ctx, msg].concat();
    |                                                        ^^^^^^ method not found in `[&[u8]; 4]`

error[E0599]: no method named `concat` found for array `[&[u8]; 4]` in the current scope
  --> /home/redacted/.cargo/registry/src/index.crates.io-6f17d22bba15001f/slh-dsa-0.1.0/src/verifying_key.rs:83:56
   |
83 |         let ctx_msg = [&[0], &ctx_len_bytes, ctx, msg].concat();
   |                                                        ^^^^^^ method not found in `[&[u8]; 4]`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `slh-dsa` (lib) due to 2 previous errors
```

It would be nice to build all crates in no_std mode in CI to detect such breakage.
Maybe we should add `cargo check --no-default-features --target thumbv7em-none-eabihf` in GitHub Actions to force no_std?

tarcieri:
Fixed in #860

PR #860, opened 2024-09-18
Remove allocation in slh-dsa

jonmon6691:
This PR takes care of some TODO's in the sign and verify functions in the slh-dsa crate, thereby resolving #854

I'm passing the new messages parts on the stack as slice references down into the functions h_msg and prf_msg so that each part can be .update()'ed into the digest without having to copy them into new memory. I think it could be improved with some type system wizardry to enforce the layout specified in FIPS 205, but that's a bit beyond my abilities.

### DIFF
--- a/slh-dsa/src/signing_key.rs
+++ b/slh-dsa/src/signing_key.rs
@@ -103,7 +103,7 @@ impl<P: ParameterSet> SigningKey<P> {
     /// Implements [slh_sign_internal] as defined in FIPS-205.
     /// Published for KAT validation purposes but not intended for general use.
     /// opt_rand must be a P::N length slice, panics otherwise.
-    pub fn slh_sign_internal(&self, msg: &[u8], opt_rand: Option<&[u8]>) -> Signature<P> {
+    pub fn slh_sign_internal(&self, msg: &[&[u8]], opt_rand: Option<&[u8]>) -> Signature<P> {
         let rand = opt_rand
             .unwrap_or(&self.verifying_key.pk_seed.0)
             .try_into()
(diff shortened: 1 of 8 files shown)


slh-dsa/src/signing_key.rs
```diff
@@ -103,7 +103,7 @@ impl<P: ParameterSet> SigningKey<P> {
     /// Implements [slh_sign_internal] as defined in FIPS-205.
     /// Published for KAT validation purposes but not intended for general use.
     /// opt_rand must be a P::N length slice, panics otherwise.
-    pub fn slh_sign_internal(&self, msg: &[u8], opt_rand: Option<&[u8]>) -> Signature<P> {
+    pub fn slh_sign_internal(&self, msg: &[&[u8]], opt_rand: Option<&[u8]>) -> Signature<P> {
```
tarcieri, line 106: This, and the `slh_verify_internal` change, are breaking. That's ok, but it would be nice if future changes weren't.

Perhaps we can at least annotate these functions with `#[doc(hidden)]` to opt out of SemVer stability for them.

jonmon6691: I'm not sure if it can be avoided, but maybe it makes more sense in the long run to move the kat's into an internal test module and then remove pub from these internal functions. Hopefully test vectors that use the "pure" and a few of the "prehash" varieties get published and those can be used as integration tests. But as far as I can tell there's nothing officially available to test the outer functions.

Let me know what you think @tarcieri and I'll make a final change for this PR

tarcieri: It's fine. I can add `#[doc(hidden)]` separately which is probably better for recording the CHANGELOG.

PR #865, opened 2024-09-29
lms: convert to hybrid-array

baloo:
Depends https://github.com/RustCrypto/hybrid-array/pull/92


lms/Cargo.toml
```diff
@@ -7,15 +7,15 @@ license = "Apache-2.0 OR MIT"
 homepage = "https://github.com/RustCrypto/signatures/tree/master/lms"
```
tarcieri, line 7: Can you also bump the version to `0.1.0-pre` to note the breaking change?

tarcieri (approved):

PR #870, opened 2024-10-12
dsa: derive `serde`

sivizius:
With this PR, it is possible to (de)serialise the `dsa`-types like `SigningKey` with `serde`.


sivizius:
Draft, because `serde_secrets` should be implemented with `serdect`.

tarcieri:
Yeah, it would be good to use `serdect`, similar to how `ecdsa` does it, especially since you're doing secret key serialization (i.e. `SigningKey`)

sivizius:
> Yeah, it would be good to use `serdect`, similar to how `ecdsa` does it, especially since you're doing secret key serialization (i.e. `SigningKey`)

I am actually working on an commit/PR which introduces another crate `common` with common types that can be used across all other crates, initially with two types: `Bytes` and `SecretBytes` which would move the zeroisation, (de)serialisisation, etc. into these type, thus harmonise (de)serialisation, etc.

tarcieri:
I'd prefer not to add superfluous crates. There shouldn't be that much duplication.

PR #871, opened 2024-10-15
workspace: limit duplicate dependencies

baloo:
This makes use of a global `patch.crates-io` instead of local `path =` attributes in the dependencies.

When pulling dependencies through git to try unreleased crates, the dependency in `path` will get duplicated. Which makes for twice the definition of types incompatible with each other.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,15 @@ members = [
 
 [profile.dev]
 opt-level = 2
+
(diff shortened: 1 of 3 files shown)


tarcieri:
Sidebar: it might be good to start using `[workspace.dependencies]` for these rather than `[patch.crates-io]`

PR #875, opened 2024-11-06
lms: fix size of `otstype`

davidthomaswood:
`otstype` is converted as a `u32` below, panics if it is `ID_LEN` (16) bytes long. With this change, the `SigningKey` can be correctly converted to bytes and back.


tarcieri:
cc @incertia

incertia (approved):
oops I just saw this. looks like a bug in the original implementation.

tarcieri (approved):

PR #877, opened 2024-11-14
Add support for ML-DSA

bifurcation:
This PR is a first step toward ML-DSA support, as discussed in #8.  It puts in place enough of the basic algebraic and encoding primitives to get key generation going.

Still TODO:
* [ ] More complete unit test coverage (e.g., on encoding routines)
* [X] Signing
* [X] Verification
* [X] Integration tests against NIST ACVP test vectors
* [X] Performance benchmarks
* [X] Pull out things that are common with ML-KEM into a shared module (?)

@tarcieri - This is currently blocked on `typenum` not having enough integers.  Specifically, there is no `ArraySize` implementation for `U4896`, which the compiler says is necessary for `MlDsa87`.  Can we get that added?

@cothan, @marija-mijailovic - If you guys are still interested in collaborating, making PRs to this branch would be a good way to help.

On the latter point about overlap with ML-KEM: Thinking about this while I was writing this PR, it seems like the main shared stuff would be:

* Structs for polynomials, polynomial vectors, NTT vectors, and NTT matrices
* Fixed-width encoding of unsigned integers

Getting those factored out would be a non-trivial savings, but would require some careful introduction of generics.  For example, the algebraic stuff would have to be generic over the field over which the polynomials exist (since there are different `q` moduli in the two algorithms), as well as the details of multiplication of polynomials and NTT.


tarcieri:
@bifurcation sure. Here's an example PR to add more sizes: https://github.com/RustCrypto/hybrid-array/pull/104

Note that the `uint!` macro uses a little endian bit encoding for the size

bifurcation:
Latest push adds signing/verification logic.  It can verify its own signatures, but fails the ACVP test vectors.

Note that it will not build against the public version of `hybrid-array`, since that is missing some `ArraySize` implementations.  I'm planning to submit a PR to that repo once things are in basic working order here, so that we have confidence that we've covered all the necessary integers.

bifurcation:
Thanksgiving update 🦃: Now passing ACVP tests, working on cleanup.

bifurcation:
@tarcieri - I think this is about ready for review and merging, though the CI won't succeed until https://github.com/RustCrypto/hybrid-array/pull/108 is merged and there's a new release.  What are the next steps here?

tarcieri:
@bifurcation new sizes released in `hybrid-array` v0.2.3

bifurcation:
OK, this is ready for review.

@tarcieri looks like the SLH-DSA build is broken at the moment.

tarcieri:
@bifurcation `slh-dsa` build is fixed if you can rebase

bifurcation:
@tarcieri - CI is green.  What do you think about review?

tarcieri:
@bifurcation sorry, have a huge backlog of PRs and haven't been reviewing as much lately. Will try to look this weekend

ml-dsa/src/lib.rs
```diff
@@ -1,2 +1,686 @@
...
+impl<P> KeyGen for P
+where
+    P: MlDsaParams,
+{
+    type KeyPair = KeyPair<P>;
+
+    /// Generate a signing key pair from the specified RNG
+    // Algorithm 1 ML-DSA.KeyGen()
+    #[cfg(feature = "rand_core")]
+    fn key_gen(rng: &mut impl CryptoRngCore) -> KeyPair<P> {
```
tarcieri, line 565: We typically use the names `generate*` or `random` for these sorts of methods. See also:

https://docs.rs/crypto-common/latest/crypto_common/trait.KeyInit.html

tarcieri (approved):
Sorry for the belated review, especially because I didn't really have much to say except this implements things the way I would've expected (e.g. using the `Signer` and `RandomizedSigner` traits).

I had some notes/bikeshed on keygen API names, but I don't see that as a reason to block merging this and getting an initial implementation landed. We can follow up on that.

tarcieri:
@bifurcation regarding extracting a shared `module_lattice` which can be reused between `ml-kem` and `ml-dsa`, I don't see that as important enough to block initial releases but a nice-to-have we can follow up on. First we'd have to figure out where to put it (possibly https://github.com/RustCrypto/utils for lack of a better option)

PR #891, opened 2025-01-23
ml-dsa: implement spki traits

### LINKED ISSUE #890: feature request: Implement spki traits on ML-DSA
Are there any plans to implement SPKI traits like spki::DynSignatureAlgorithmIdentifier and spki::EncodePublicKey so that x509_cert::builder can  be used?

Thanks!

baloo:
RFC Draft: [https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/](https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/)

Fixes #890


ml-dsa/src/lib.rs
```diff
@@ -156,6 +191,23 @@ impl<P: MlDsaParams> signature::KeypairRef for KeyPair<P> {
     type VerifyingKey = VerifyingKey<P>;
 }
 
+impl<P> TryFrom<PrivateKeyInfoRef<'_>> for KeyPair<P>
```
baloo, line 194: I'd want to implement `EncodePrivateKey for KeyPair` but I couldn't find a way to recover the seed from a signing key.

baloo:
cc @ArthurHeymans
I did not throw that in with x509_cert::builder quite yet.

ml-dsa/src/lib.rs
baloo: Because seed -> privatekey is a one way function. We'd have to keep the seed in the keypair if we wanted to do that.

baloo: https://github.com/RustCrypto/signatures/pull/892

ml-dsa/src/lib.rs
```diff
@@ -384,8 +436,34 @@ impl<P: MlDsaParams> signature::RandomizedSigner<Signature<P>> for SigningKey<P>
...
+}
+
+impl<P> TryFrom<PrivateKeyInfoRef<'_>> for SigningKey<P>
+where
+    P: MlDsaParams,
+{
+    type Error = pkcs8::Error;
+
+    fn try_from(private_key_info: pkcs8::PrivateKeyInfoRef<'_>) -> pkcs8::Result<Self> {
+        private_key_info
```
baloo, line 453: Todo: change that to just consume Keypair

baloo:
I've pulled this and #892 in a playground: https://github.com/baloo/playground-ml-dsa

ml-dsa/Cargo.toml
```diff
@@ -15,14 +15,17 @@ categories = ["cryptography"]
 keywords = ["crypto", "signature"]
 
 [features]
-default = ["std", "rand_core"]
-std = ["sha3/std"]
+default = ["std", "rand_core", "alloc"]
+std = ["alloc", "sha3/std"]
```
tarcieri, line 19: Sidebar: what is this even for? I don't see it used anywhere in the code

ml-dsa/Cargo.toml
```diff
@@ -15,14 +15,17 @@ categories = ["cryptography"]
...
 zeroize = ["dep:zeroize", "hybrid-array/zeroize"]
 rand_core = ["dep:rand_core", "signature/rand_core"]
+alloc = ["pkcs8/alloc"]
 
 [dependencies]
+const-oid = { version = "0.10.0-rc.1", features = ["db"] }
 hybrid-array = { version = "0.2.3", features = ["extra-sizes"] }
 num-traits = "0.2.19"
+pkcs8 = { version = "=0.11.0-rc.1", default-features = false }
```
tarcieri, line 28: Elsewhere we don't make these hard dependencies. Perhaps make them optional with a `pkcs8 = ["dep:const-oid", "dep:pkcs8"]` feature? They can be on-by-`default`.

ml-dsa/Cargo.toml
tarcieri: Opened a PR to remove it: https://github.com/RustCrypto/signatures/pull/893

baloo:
rebased on https://github.com/RustCrypto/signatures/pull/893

ml-dsa/Cargo.toml
```diff
@@ -15,9 +15,11 @@ categories = ["cryptography"]
 keywords = ["crypto", "signature"]
 
 [features]
-default = ["rand_core"]
+default = ["rand_core", "alloc", "pkcs8"]
 zeroize = ["dep:zeroize", "hybrid-array/zeroize"]
 rand_core = ["dep:rand_core", "signature/rand_core"]
+alloc = ["pkcs8?/alloc"]
+pkcs8 = ["const-oid", "dep:pkcs8"]
```
tarcieri, line 22:
```suggestion
pkcs8 = ["dep:const-oid", "dep:pkcs8"]
```

PR #892, opened 2025-01-24
ml-dsa: encode private key

baloo:
Follow up on #891

This reworks the `ml_dsa::KeyPair` to keep the seed along with the `SigningKey` which allows to serialize the private key to the pkcs8 format after that.


tarcieri:
@baloo can you rebase now that I've merged #893?

tarcieri:
One more rebase and I think this is good to go

baloo:
Yeah I’ll do that in 2-3h when I get back home.

tarcieri (approved):

PR #895, opened 2025-02-12
Make ML-DSA signature decoding follow the spec

### LINKED ISSUE #894: ml-dsa: some valid signature values don't successfully round trip through encode + decode
I'm hitting this issue when testing v0.1.0-pre.0 from crates.io and and when testing git master.

The failure happens when calling `Signature::<MlDsa65>::decode()`.  Lot's of valid signatures fail (like ~1% ish), so problematic ones aren't hard to find. Here's an example value that is failing for me:

```rust
// Minimal demonstration of a valid signature that wont round-trip through encode/decode.

use ml_dsa::{EncodedSignature, KeyGen, MlDsa65, Signature, B32};
use signature::Signer;

static SEED: [u8; 32] = [
    197, 185, 159, 59, 216, 233, 208, 40, 244, 4, 182, 73, 109, 244, 205, 113, 116, 55, 206, 145,
    214, 205, 247, 130, 41, 113, 93, 14, 140, 194, 191, 232,
];
static MESSAGE: &[u8] = b"There seems to be a round tripping issue somewhere in here";

fn main() {
    let mut seed = B32::default();
    seed.0.copy_from_slice(&SEED);
    let kp = MlDsa65::key_gen_internal(&seed);
    let sig = kp.signing_key().sign(MESSAGE);

    let mut sig_buf = [0; 3309];
    sig_buf.copy_from_slice(sig.encode().as_slice());
    let sig_enc = EncodedSignature::<MlDsa65>::try_from(&sig_buf[..]).unwrap();
    let _sig = Signature::<MlDsa65>::decode(&sig_enc).unwrap(); // This will fail
}
```

Further code demonstrating the bug is here: https://github.com/json420/ml_dsa_bug

bifurcation:
Fixes #894 

This PR also adds a test that runs many random signatures.  I confirmed locally that this produced a failure rate of 0.4-1.1% for ML-DSA-65 and ML-DSA-87, all due to signature decoding, and all for the reasons discussed below.  I never saw any failures for ML-DSA-44, but that seems plausible: The failure condition arises when the high-order bits of the `r` vector don't change, and the threshold $\gamma_2 / 2$ is roughly a factor of 2.75 lower for ML-DSA-44 than the other variants.

It appears that the root of the problem is a one-character fix, allowing repeated values in the monotonicity check in `hint.rs`.  This makes it so that the tests usually pass.  Marking this PR as draft because that same test fails intermittently, reporting `internal error: entered unreachable code` at `hint.rs:30`.  I'll take a look at that tomorrow.

### DIFF
--- a/ml-dsa/src/lib.rs
+++ b/ml-dsa/src/lib.rs
@@ -927,7 +927,7 @@ mod test {
             let sig_dec = Signature::<P>::decode(&sig_enc).unwrap();
 
             assert_eq!(sig_dec, sig);
-            assert!(vk.verify_internal(&[M], &sig));
+            assert!(vk.verify_internal(&[M], &sig_dec));
         }
     }
 
--- a/ml-dsa/src/lib.rs
+++ b/ml-dsa/src/lib.rs
@@ -1,4 +1,4 @@
-#![no_std]
+// XXX #![no_std]
 #![doc = include_str!("../README.md")]
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
@@ -89,7 +89,7 @@ pub use crate::util::B32;
 pub use signature::Error;
 
 /// An ML-DSA signature
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, Debug)]
 pub struct Signature<P: MlDsaParams> {
     c_tilde: Array<u8, P::Lambda>,
     z: Vector<P::L>,
@@ -898,5 +898,63 @@ mod test {
         sign_verify_round_trip_test::<MlDsa44>();
         sign_verify_round_trip_test::<MlDsa65>();
         sign_verify_round_trip_test::<MlDsa87>();
+    }
+
+    fn many_round_trip_test<P>()
+    where
+        P: MlDsaParams,
+    {
+        use rand::Rng;
+
+        const ITERATIONS: usize = 1000;
+
+        let mut rng = rand::thread_rng();
+        let mut seed = B32::default();
+
+        for _i in 0..ITERATIONS {
+            let seed_data: &mut [u8] = seed.as_mut();
+            rng.fill(seed_data);
+
+            let kp = P::key_gen_internal(&seed);
+            let sk = kp.signing_key;
+            let vk = kp.verifying_key;
+
+            let M = b"Hello world";
+            let rnd = Array([0u8; 32]);
+            let sig = sk.sign_internal(&[M], &rnd);
+
+            let sig_enc = sig.encode();
+            let sig_dec = Signature::<P>::decode(&sig_enc).unwrap();
+
+            assert_eq!(sig_dec, sig);
+            assert!(vk.verify_internal(&[M], &sig));
+        }
+    }
+
+    #[test]
+    fn many_round_trip() {
+        many_round_trip_test::<MlDsa44>();
+        many_round_trip_test::<MlDsa65>();
+        many_round_trip_test::<MlDsa87>();
+    }
+
+    #[test]
+    fn encode_decode_fail() {
+        use signature::Signer;
+
+        const SEED: [u8; 32] = [
+            197, 185, 159, 59, 216, 233, 208, 40, 244, 4, 182, 73, 109, 244, 205, 113, 116, 55,
+            206, 145, 214, 205, 247, 130, 41, 113, 93, 14, 140, 194, 191, 232,
+        ];
+        const MESSAGE: &[u8] = b"There seems to be a round tripping issue somewhere in here";
+
+        let mut seed = B32::default();
+        seed.0.copy_from_slice(&SEED);
+
+        let seed = SEED.into();
+        let kp = MlDsa65::key_gen_internal(&seed);
+        let sig = kp.signing_key().sign(MESSAGE);
+        let sig_enc = sig.encode();
+        let _sig = Signature::<MlDsa65>::decode(&sig_enc).unwrap();
     }
 }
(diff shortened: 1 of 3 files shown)


bifurcation:
Heh, and apparently I have some clippy clean-up to do as well!

ml-dsa/src/lib.rs
```diff
@@ -899,4 +899,62 @@ mod test {
...
+            let rnd = Array([0u8; 32]);
+            let sig = sk.sign_internal(&[M], &rnd);
+
+            let sig_enc = sig.encode();
+            let sig_dec = Signature::<P>::decode(&sig_enc).unwrap();
+
+            assert_eq!(sig_dec, sig);
+            assert!(vk.verify_internal(&[M], &sig));
+        }
+    }
```
tarcieri, line 932: Perhaps consider `proptest` for these sorts of tests.

You can look at https://github.com/RustCrypto/crypto-bigint/tree/master/tests for some examples.

(Or I can try to convert them to proptests as a followup)

json420:
Sorry, didn't realize this was ready to test already.

Still having problems, but with a new error:

thread 'ownedchain::tests::test_ownedchainstore' panicked at /home/jderose/.cargo/git/checkouts/signatures-7e4d4cb41d25b54f/92d2a82/ml-dsa/src/hint.rs:30:9:
internal error: entered unreachable code

bifurcation:
Yep, that looks like the same error I'm seeing locally.  Still working on that one.

bifurcation:
@json420 Oddly, when I went to fix that remaining `unreachable!()` bug this morning, I was unable to reproduce it.  It failed the first few times I tried the test, but then never after, despite hundreds of thousands of trials.  Are you still seeing it?

json420:
@bifurcation Huh, I hit what I pasted right away when I first ran my tests against your branch... but now I'm not getting them anymore. I'll test further.

json420:
Ok, I can still hit the error (sometimes) in my tests:

thread 'ownedchain::tests::test_ocs_secret_to_public' panicked at /home/jderose/.cargo/git/checkouts/signatures-7e4d4cb41d25b54f/f1e4e30/ml-dsa/src/hint.rs:30:9:
internal error: entered unreachable code

I'll work on narrowing it down to known inputs.

json420:
@bifurcation I still don't have reproduce-able values, but I have a theory: "internal error: entered unreachable code" is only being hit on legitimately invalid signature values. I've ran through millions of valid signatures and haven't hit this, but one difference in my zebrachain tests is that I'm testing bit-flip permutations, which in some cases should produce a signature that shouldn't decode (right?).

tarcieri:
Perhaps you can change the `unreachable!` to print out the signature when it’s hit?

bifurcation:
@json420 - Bit flips could certainly cause the decoding to fail, but that `unreachable!()` should actually be unreachable.  So I would appreciate it if you could send a signature that's tripping it and I can dig into it.

@tarcieri - Should we consider this `unreachable!()` issue an independent problem, and go ahead and merge this fix?  I am also reading up on `proptest`, per your suggestion above, but would propose handling that as a follow-on.

bifurcation:
Actually @tarcieri please hold, it started failing for me again.

bifurcation:
This test case reliably fails:

``` rust
    #[test]
    fn bad_seed() {
        const BAD_SEED: &[u8] = &[
            223, 116, 86, 28, 89, 196, 31, 159, 108, 178, 58, 1, 14, 198, 242, 131, 243, 77, 188,
            21, 25, 108, 151, 27, 232, 208, 147, 229, 245, 113, 178, 114,
        ];

        let seed = B32::try_from(BAD_SEED).unwrap();

        let kp = MlDsa87::key_gen_internal(&seed);
        let sk = kp.signing_key;
        let vk = kp.verifying_key;

        let M = b"Hello world";
        let rnd = Array([0u8; 32]);
        let sig = sk.sign_internal(&[M], &rnd);

        let sig_enc = sig.encode();
        let sig_dec = Signature::<MlDsa87>::decode(&sig_enc).unwrap();
        let _ver = vk.verify_internal(&[M], &sig); // <-- this will panic
    }
```

json420:
@bifurcation yup, that one fails for me too. Also fails against git master..

ml-dsa/src/lib.rs
tarcieri: Note: tried doing this myself and it's a tad tricky due to the use of generics

tarcieri: Okay, found an okay enough solution and opened a PR to add proptests: https://github.com/RustCrypto/signatures/pull/896

It also required adding a `Debug` impl for `Signature` as this PR does also.

It was able to find some failures right away, though I did see `many_round_trip_test` find failures with this PR's current state that the proptests did not (but perhaps they just need to be run more).

bifurcation:
OK, I believe the latest commit fixes this secondary issue.  As with the first issue, we were being too strict in our assumptions about what ranges of numbers would appear in a given setting.  The `r0` value that caused the panic was assumed to be in the range `(-gamma2, gamma2]` produced by `mod_plus_minus()`.  However, `r0` is actually produced by `decompose()`, which sometimes subtracts one from the `r0` output.  So the actual range of possible `r0` outputs from `decompose()` is `[-gamma2, gamma2]` (including `-gamma2` on the low end).

Once again, the actual fix is one character, changing `>` to `>=`.  In the process of debugging, I added tests verifying the correctness of `mod_plus_minus()` and `decompose()`, which I have left in for posterity.  These tests are good candidates for proptests, cf. #896.

@json420 - Could you confirm that this fix looks good to you?

json420:
@bifurcation I haven't been able to get any failures since your last change. Thanks!

bifurcation:
OK @tarcieri I think this is ready to merge.

PR #896, opened 2025-02-12
Add proptests for `Signature` encoding

tarcieri:
Ensures encoded signatures can be round tripped successfully, i.e. encoded and then decoded again without errors or panics.


tarcieri:
Closing in favor of #897, which is more comprehensive

PR #897, opened 2025-02-12
Add proptests with round trip check

tarcieri:
- Ensures we can verify signatures we've produced
- Ensures we can decode signatures we've encoded

Note: need to land #895 first so the tests pass


ml-dsa/src/lib.rs
```diff
@@ -192,6 +192,11 @@ impl<P: MlDsaParams> KeyPair<P> {
     pub fn signing_key(&self) -> &SigningKey<P> {
         &self.signing_key
     }
+
+    /// The verifying key of the key pair
+    pub fn verifying_key(&self) -> &VerifyingKey<P> {
```
tarcieri, line 197: I noticed there wasn't an inherent method for this and it was being accessed exclusively through `signature::Keypair` which I found a little surprising.

ml-dsa/tests/proptests.rs
```diff
@@ -0,0 +1,51 @@
...
+       MlDsa65::key_gen_internal(seed_bytes.as_ref())
+    }
+}
+prop_compose! {
+    fn mldsa87_keypair()(seed_bytes in any::<[u8; 32]>()) -> KeyPair<MlDsa87> {
+        MlDsa87::key_gen_internal(seed_bytes.as_ref())
+    }
+}
+
+macro_rules! round_trip_test {
```
tarcieri, line 24: This needs to be a macro as opposed to a function to use `prop_assert_eq!`

ml-dsa/tests/proptests.proptest-regressions
```diff
@@ -0,0 +1,8 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
```
bifurcation, line 6: I don't love having this gigantic, un-interpretable file in the repo.  Can it get pruned?  Maybe we could reset it after we merge the known issues in #895.

bifurcation (approved):
Overall, looks good.  Couple of minor comments.

ml-dsa/tests/proptests.proptest-regressions
tarcieri: It's SOP for proptest to check in these files. See description here: https://altsysrq.github.io/proptest-book/proptest/getting-started.html

> If we look at the top directory after the test fails, we’ll see a new proptest-regressions directory, which contains some files corresponding to source files containing failing test cases. These are [failure persistence](https://altsysrq.github.io/proptest-book/proptest/failure-persistence.html) files. The first thing we should do is add these to source control.

It's a log of every failure that has been found in the code so far so you automatically get regression tests for all of them.

ml-dsa/tests/proptests.rs
```diff
@@ -0,0 +1,51 @@
...
+    ($params:path, $keypair:expr) => {
+        let sig = $keypair.signing_key().sign_internal(&[MSG], RND.as_ref());
+
+        // Ensure verification doesn't crash
+        let _ = $keypair.verifying_key().verify_internal(&[MSG], &sig);
```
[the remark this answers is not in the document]
tarcieri: Added a check for the verification result

ml-dsa/tests/proptests.rs
```diff
@@ -0,0 +1,51 @@
+use ml_dsa::{KeyGen, KeyPair, MlDsa44, MlDsa65, MlDsa87, Signature};
+use proptest::prelude::*;
+
+const MSG: &[u8] = b"Hello world";
+const RND: [u8; 32] = [0u8; 32];
```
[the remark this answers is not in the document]
tarcieri: Switched to using the `Signer` and `Verifier` traits. This was just cargo culted from your original randomized tests.

PR #904, opened 2025-02-16
ed25519: bump `rand_core` to `0.9.0`

baloo:


json420:
As far as I can tell this is correct (but I'm new here, haha).

I'd love to see this move forward because this is needed to be able to use `ml-dsa` and `e25519` together again.

In local testing, I can `cargo add ed25519` from this branch, then `cargo add ml-dsa` from git master, and it plays nice as long as I add ml-dsa with `--no-default-features`.

If the ml-dsa `pkcs8` feature is enabled, things blow up because ml-dsa is currently locked to `pkcs8=0.11.0-rc.1`. That's not a problem with this PR, just something I thought I'd note that needs fixing.

Anyway, I give this PR a 👍

baloo:
closed in favor of #913

PR #906, opened 2025-02-16
dsa: migrate to `crypto-bigint`

baloo:
This is a rebase of #784

There are two commits on top that I think should fix tests. There is a bit of cleanup to do still.


baloo:
cc @aumetra

aumetra:
Apologies, I was a bit absent lately. I'll have a look when I get the time (which should be in like two hours at max)

baloo:
No need to apologize! Please take care.
That was only to keep you updated on the progress and let you chime in as needed.

If there are specific things you wanted to take care of, please let me know I'd be happy to.

baloo:
out of the last 3 tests that still fail:
 - sign_message: that one needs the vectors to be adjusted, we're switching the backend and by doing so the primes we get from the RNG are different because their generation is now different.
 - signer_verifier_signature: that one is a precision that needs to be adjusted.
 - verify_signature: I'm not quite sure yet.

aumetra:
> Overall having to adjust precisions seems to be ... inefficent? error-prone? I guess there is a way better method for doing that. I need to have a look at RSA how they did it there.

Seems like even `rsa` did the same thing. Just via the `.widen` method which is, I guess, a newer method that didn't exist when I started this PR a year ago

aumetra:
Scratch that, it existed, I just didn't see it at that time

baloo:
okay, it's definitely on the ugly side, but tests do pass \o/

I think I got the hang of it, I think I can clean things up from there, but ... tomorrow.

aumetra:
Nice! I mean I have a few hours to kill today, too, so I can go ahead and try to clean up some things, too. I'll PR it into your fork when I get to it

dsa/src/signing_key.rs
```diff
@@ -80,31 +87,57 @@ impl SigningKey {
...
+        let params = BoxedMontyParams::new(p.clone());
+        let form = BoxedMontyForm::new((**g).clone(), params);
+        let r = form.pow(&k).retrieve() % q.widen(p.bits_precision());
+        debug_assert_eq!(key_size.l_aligned(), r.bits_precision());
+
+        let r_short = r.shorten(key_size.n_aligned());
+        // TODO(baloo): is there any way R could be zero here? Could it be any reason for K to be
+        //              one?
+        let r_short = NonZero::new(r_short).unwrap();
+        let r = NonZero::new(r).unwrap();
```
baloo, line 124: I'm not sure about this unwrap, isn't there a way for `r = g.modpow(k, p) % q` to be zero?

dsa/src/signing_key.rs
```diff
@@ -80,31 +87,57 @@ impl SigningKey {
...
-        let z = BigUint::from_bytes_be(&hash[..z_len]);
+        let z_len = min(n as usize, block_size);
+        let z = BoxedUint::from_be_slice(&hash[..z_len], z_len as u32 * 8)
+            .expect("invariant violation");
 
-        let s = (inv_k * (z + x * &r)) % q;
+        let s = inv_k.mul_mod(&(z + &**x * &*r), &q.widen(key_size.l_aligned()));
+        let s = s.shorten(key_size.n_aligned());
+        // TODO(baloo): is there any way S could be zero here?
+        let s = NonZero::new(s).unwrap();
```
baloo, line 136: Similarly here, s could be zero?

baloo (commented):
Those two `NonZero` I'm not sure about.

dsa/tests/signature.rs
```diff
@@ -22,19 +23,19 @@ const MESSAGE: &[u8] = b"test";
...
-const MESSAGE_SIGNATURE_OPENSSL_ASN1: &[u8] = &[
-    0x30, 0x2C, 0x02, 0x14, 0x6D, 0xB3, 0x8E, 0xAF, 0x97, 0x13, 0x7E, 0x07, 0xFF, 0x24, 0xB8, 0x66,
-    0x97, 0x18, 0xE1, 0x6F, 0xD7, 0x9A, 0x28, 0x2D, 0x02, 0x14, 0x47, 0x8C, 0x0B, 0x96, 0x51, 0x08,
-    0x08, 0xC8, 0x34, 0x9D, 0x0D, 0x41, 0xC7, 0x73, 0x0F, 0xB5, 0x9C, 0xBB, 0x00, 0x34,
-];
+const MESSAGE_SIGNATURE_OPENSSL_ASN1: &[u8] = &hex!(
+    "302e 0215 0092 4523 9462 f59b 5936 d355
+     8b16 8dbb 0915 7ba1 5302 1500 9d8d 7aa7
+     02a6 2dde 1975 e9f2 a20c 73b4 7dd3 0912"
+);
```
baloo, line 38: The behavior of `Components::generate` changed because of the switch to crypto-bigint (and crypto-primes). This makes the "deterministic" private key used for test change. And as a result, makes the signatures change.

I don't believe we offered any guarantee towards the stability of the `Component::generate`?

aumetra: No, we didn't make any guarantees. The only guarantee we provided was "it outputs something that works".  

If some code relies on this staying the same it's really fragile anyway, since they'd have to use the same algorithms to generate primes and factors, which can differ between implementations. So yeah, this should be fine.

dsa/src/signing_key.rs
baloo: Oh right, that ties to the `k` value computed in the `secret_number` and `secret_number_rfc6979`!

That would then be a bug and a panic might be in order.

baloo:
This probably should include tests stretching the bit precision of a signature to test against panic.
Maybe proptests would be a good fit?

tarcieri:
Yes proptests would be good.

I need to pull this down and benchmark it but getting the precisions right has been a challenge in `rsa`. Did performance regress significantly?

baloo:
It was pretty bad at some point at some point (tests would take over a minute) but I did not dig why. It works fine now.

I did not run a benchmark.

dignifiedquire:
anything missing for merging this?

tarcieri:
cc @baloo

baloo:
A review!

dsa/src/generate/secret_number.rs
```diff
@@ -2,32 +2,16 @@
...
-    let mut reduced = (hash % q).to_bytes_be();
-
-    while reduced.len() < q_byte_len {
-        reduced.insert(0, 0);
-    }
-
-    reduced
+fn reduce_hash(hash: &[u8], desired_size: usize) -> &[u8] {
+    &hash[(hash.len() - desired_size)..]
 }
```
tarcieri, line 15: What happened here?

dsa/src/lib.rs
```diff
@@ -182,8 +190,8 @@ impl TryFrom<&[u8]> for Signature {
...
 }
 
-/// Returns a `BigUint` with the value 2
+/// Returns a `BoxedUint` with the value 2
 #[inline]
-fn two() -> BigUint {
-    BigUint::from(2_u8)
+fn two() -> BoxedUint {
+    BoxedUint::from(2_u8)
 }
```
tarcieri, line 197: A `Uint` constant could possibly work here, depending on usage

dsa/src/verifying_key.rs
```diff
@@ -56,23 +65,47 @@ impl VerifyingKey {
...
             return Some(false);
         }
 
-        let w = s.mod_inverse(q)?.to_biguint().unwrap();
+        let q = q.widen(p.bits_precision());
+        let q = &q;
+        let r = r.widen(p.bits_precision());
+        let r = &r;
+        let s = s.widen(p.bits_precision());
+        let s = &s;
```
tarcieri, line 77: This is just to save writing `&q`, `&r`, and `&s` I guess?

tarcieri:
As a general comment there's a lot of usage of `expect` in places that it wasn't required with `num-bigint`.

I don't want to make addressing those a blocker for this PR but it sure would be nice to fix eventually, even if it's just converting them to errors internally.

tarcieri:
Hmm, what's up with the test failures?

baloo:
> Hmm, what's up with the test failures?

Tests are relying on the generation of primes by the inner library (injecting a rng). Whenever the implementation for primes changes, the tests will fail. I just bumped the pin for `crypto-primes` and it got reworked a bit. I'll need to adjust the expected signatures.

dsa/src/lib.rs
baloo: This is used to compare against components and those are BoxedUint, I would need an `impl PartialOrd<Uint> for BoxedUint` for that.

tarcieri: Seems doable

dsa/src/generate/keypair.rs
```diff
@@ -3,16 +3,25 @@
...
+            let x = BoxedUint::random_mod(rng, components.q());
+            if let Some(x) = NonZero::new(x).into() {
+                break 'gen_x x;
+            }
+        };
+
+        if let Some(y) = components::public(&components, &x).into_option() {
+            break (x, y);
+        }
+    };
```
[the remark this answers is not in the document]
baloo: let me know what you think.

dsa/src/verifying_key.rs
baloo: Just me when I was fiddling with precisions.

dsa/src/generate/secret_number.rs
baloo: The logic moved in `secret_number_rfc6979`, the name was incorrect, that should have been `truncate_hash` or something.

.github/workflows/dsa.yml
```diff
@@ -17,14 +17,15 @@ env:
 
 jobs:
   build:
+    if: false # both thumbv7 and wasm require a getrandom backend to be injected
```
tarcieri, line 20: Why is `getrandom` a hard dependency now?

dsa/src/generate/keypair.rs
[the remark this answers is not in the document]
tarcieri: I can't tell what changed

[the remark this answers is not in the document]
baloo: https://github.com/RustCrypto/signatures/pull/906/commits/87db89af3996e71fa77491f1e48b8930ea2a3596

.github/workflows/dsa.yml
baloo: wait? is it?

tarcieri: Offhand I don't see anything which would make `getrandom` a hard dependency so I guess I'm confused why this is needed

baloo: yeah you're correct, I guess I got confused with the other rand_core upgrade? I don't remember.

tarcieri:
@baloo there are some merge conflicts

tarcieri (approved):

PR #910, opened 2025-02-20
ml-dsa: update signature@2.3.0-pre.6

json420:
Tiny change, but... my first for RustCrypto stuffs, so I'm super excited.

As far as I can tell, this is the only change need to make ml-dsa build against signatures@2.3.6-pre.6.

Have an awesome day and happy Rust 1.85! 💖 🦀

### DIFF
--- a/ml-dsa/Cargo.toml
+++ b/ml-dsa/Cargo.toml
@@ -26,7 +26,7 @@ hybrid-array = { version = "0.2.3", features = ["extra-sizes"] }
 num-traits = "0.2.19"
 rand_core = { version = "0.9", optional = true }
 sha3 = "0.10.8"
-signature = "=2.3.0-pre.5"
+signature = "2.3.0-pre.6"
 zeroize = { version = "1.8.1", optional = true, default-features = false }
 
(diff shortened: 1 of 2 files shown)


json420:
Although I just noticed I drop the = in the version. I guess this may or may not be the right thing base on release plans for traits/signature.

ml-dsa/Cargo.toml
```diff
@@ -26,7 +26,7 @@ hybrid-array = { version = "0.2.3", features = ["extra-sizes"] }
 num-traits = "0.2.19"
 rand_core = { version = "0.9", optional = true }
 sha3 = "0.10.8"
-signature = "=2.3.0-pre.5"
+signature = "2.3.0-pre.6"
```
tarcieri, line 29: Please retain `=` because there are breaking changes from release-to-release (really `=` should be the default for prereleases)

json420: @tarcieri thanks for the clarification on the "=". I'm still getting hang for how things move in the RustCrypto world ;D

Updated my branch.

PR #911, opened 2025-02-22
ml-dsa: update pkcs8@0.11.0-rc.2

json420:
As far as I can tell, this is the only change need to make ml-dsa build against pkcs8 0.11.0-rc.2.

### DIFF
--- a/ml-dsa/Cargo.toml
+++ b/ml-dsa/Cargo.toml
@@ -30,13 +30,13 @@ signature = "=2.3.0-pre.6"
 zeroize = { version = "1.8.1", optional = true, default-features = false }
 
 const-oid = { version = "0.10.0-rc.1", features = ["db"], optional = true }
-pkcs8 = { version = "=0.11.0-rc.1", default-features = false, optional = true }
+pkcs8 = { version = "=0.11.0-rc.2", default-features = false, optional = true }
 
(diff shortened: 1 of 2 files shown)


ml-dsa/Cargo.toml
```diff
@@ -30,13 +30,13 @@ signature = "=2.3.0-pre.6"
 zeroize = { version = "1.8.1", optional = true, default-features = false }
 
 const-oid = { version = "0.10.0-rc.1", features = ["db"], optional = true }
-pkcs8 = { version = "=0.11.0-rc.1", default-features = false, optional = true }
+pkcs8 = { version = "=0.11.0-rc.2", default-features = false, optional = true }
```
baloo, line 33: I think the better approach would be to remove the =
```suggestion
pkcs8 = { version = "0.11.0-rc.2", default-features = false, optional = true }
```

similarly below.

We only really need to be bind a precise version for pre-releases, we should be more lose for rc releases. Unless you're using a specific api released in `0.11.0-rc.2` relying on the dependency resolver should work (and it's more future-proof).

json420: Thanks, that makes sense! Changed!

baloo (approved):

PR #913, opened 2025-02-23
Upgrade all crates to the 2024 edition and bump MSRV to 1.85

baloo:
This currently excludes:
 - dsa because it's getting ported over to crypto-bigint
 
 This depends on:
  - https://github.com/RustCrypto/traits/pull/1751
  - https://github.com/RustCrypto/traits/pull/1767
  - https://github.com/RustCrypto/crypto-bigint/pull/762
  - https://github.com/RustCrypto/crypto-bigint/pull/765
  - https://github.com/zkcrypto/ff/pull/122
  - https://github.com/zkcrypto/group/pull/56
  - https://github.com/RustCrypto/hashes/pull/652
  - https://github.com/RustCrypto/MACs/pull/178


lms/src/ots/private.rs
```diff
@@ -107,7 +107,7 @@ impl<Mode: LmsOtsMode> RandomizedSignerMut<Signature<Mode>> for SigningKey<Mode>
 
         // Generate the message randomizer C
         let mut c = <Output<Mode::Hasher>>::default();
-        rng.fill_bytes(&mut c);
+        rng.try_fill_bytes(&mut c).map_err(|_| Error::new())?;
```
baloo, line 110: We kind of eat the error here. This might be a bad idea.

I wish we could do Error::from_source instead, but that would require the RandomizedSignerMut to change its signature to something like:
```
fn try_sign_with_rng<E: core:error::Error, R: TryCryptoRng<Error = E>>(
        &mut self,
        rng: &mut R,
        msg: &[u8],
    ) -> Result<Signature<Mode>, Error> {
``
Not sure.

File: ecdsa/src/signing.rs [hunk 2]
```diff
@@ -197,14 +195,14 @@ where
...
+    fn sign_prehash_with_rng<R: TryCryptoRng + ?Sized>(
         &self,
-        rng: &mut impl CryptoRngCore,
+        rng: &mut R,
         prehash: &[u8],
     ) -> Result<Signature<C>> {
         let z = bits2field::<C>(prehash)?;
         let mut ad = FieldBytes::<C>::default();
-        rng.fill_bytes(&mut ad);
+        rng.try_fill_bytes(&mut ad).map_err(|_| Error::new())?;
```
baloo, line 205: similar situation here.

baloo (commented):
Opinion about `RandomizedSignerMut` signature?

lms/src/ots/private.rs
tarcieri: We can potentially consider something like that

tarcieri:
@baloo can you rebase?

erik-3milabs:
Would it be possible to include new `elliptic-curve` and `ecdsa` pre-releases with this PR? :smile: 

Currently, `ecdsa@0.17.0-pre.9`  and `crypto-bigint@0.6.0` do not work together, because:
- `elliptic-curve@0.14.0-rc.1` (on which `ecdsa` depends) pins `serdect=0.3.0-rc.0`, while
- `crypto-bigint@0.6.0` requires `^0.3`. 

`elliptic-curve` [has already been updated](https://github.com/RustCrypto/traits/commit/3597746976a972960152c1b562000b44d5a69da6) to use `0.3.0`; it just needs a new release for these packages to work together again.

I might be mistaken, but all I think needs to happen is:
- release `elliptic-curve@0.14.0-rc.2`
- have `ecdsa` depend on this new release, and
- release `ecdsa@0.17.0-pre.9`.

If this is not the right place and/or time, I'm happy to create a separate issue for it :)

tarcieri:
@erik-3milabs we generally do releases in separate PRs to keep ones like this focused

baloo:
Just to mitigate expectations, I don't know if we'll be able to release elliptic-curve until the situation of `group` and `ff` support of `rand_core 0.9` is sorted out.

erik-3milabs:
> @erik-3milabs we generally do releases in separate PRs to keep ones like this focused

In that case, would there be a point in raising an issue for this?

tarcieri:
Sure, please open a separate issue

baloo:
Anything else needed here?

tarcieri (approved):

PR #915, opened 2025-02-26
ecdsa: provide a SigningKey::try_from_rng

baloo:
Depends:
 - https://github.com/zkcrypto/ff/pull/126
 - https://github.com/zkcrypto/ff/pull/127
 - https://github.com/RustCrypto/traits/pull/1774


tarcieri:
@baloo rebase?

baloo:
I think that should merge fine and the failing test should be fixed by #923
I can rebase in an hour otherwise.

Issue #887, opened 2025-01-21
ml-dsa: large crate size due to test vectors


tarcieri:
I just noticed [`ml-dsa` v0.1.0-pre.0](https://crates.io/crates/ml-dsa/0.1.0-pre.0) is 1.44MB in size due to the `.json` files for KATs/test vectors.

Perhaps we could omit these from the published crate to reduce its size?

cc @bifurcation

json420:
I think removing the KAT files is reasonable, but if there ever is a way to `cargo test` on all dependencies as well, as discussed here:

https://internals.rust-lang.org/t/a-way-to-run-tests-for-all-your-dependencies-recursively/20584

Then perhaps a `ml-dsa-kat` package could be added with the data files and any additional test code. Dependents could add this to their dev dependencies. That's me thinking in Debian packaging terms though, so not sure if that's the correct approach in Rust land.

bifurcation:
Yeah, I don't see a reason to package the tests in distributions.  They're in the repo purely for development purposes; I wouldn't expect anyone importing the crate to use them.  But I'm definitely not an expert in crate packaging.

tarcieri:
Fixed in #919

PR #919, opened 2025-03-05
ml-dsa: exclude tests/ in published crate due to size

json420:
This addresses bug #887 "ml-dsa: large crate size due to test vectors".

The simplest solution is to exclude all of tests/ in the published crate, so that's what my pull request does.

### DIFF
--- a/ml-dsa/Cargo.toml
+++ b/ml-dsa/Cargo.toml
@@ -13,6 +13,7 @@ repository = "https://github.com/RustCrypto/signatures"
 readme = "README.md"
 categories = ["cryptography"]
 keywords = ["crypto", "signature"]
+exclude = ["tests"]
 
 [features]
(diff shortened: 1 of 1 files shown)


tarcieri:
I think it would be good to just exclude the tests which rely on the large JSON files, i.e. `key-gen`, `sig-gen`, and `sig-ver`

tarcieri:
Sure, that's fine

json420:
Okay, I updated my pull request to exclude the 3 json files. I also excluded their corresponding .rs files, otherwise those tests fail.

Finally, I'm excluding `tests/proptests.proptest-regressions`. Removing it doesn't seem to cause any test  failures.

tarcieri (approved):

tarcieri:
Thanks!

PR #921, opened 2025-03-08
ml-dsa: catch `benches/ml_dsa.rs` up with ML-DSA implementation changes

pqcfox:
Currently, when running `cargo bench` in `ml-dsa`, a series of compile warnings and errors result as `benches/ml_dsa.rs` is out of sync with the ML-DSA implementation.

This PR updates `benches/ml_dsa.rs` to eliminate these compile warnings and errors, returning the benchmarking code to a working state.


json420:
With this fix, cargo bench compiles and runs fine for me :+1: 

It would be handy to have a .github/workflow to ensure the benchmarks are always in running order.

json420:
Testing with `cargo bench --no-run` might a good approach... it makes sure the benchmarks compile, but doesn't actually run them (in case any of them are too time consuming for a github workflow).

tarcieri:
We usually use `cargo build --benches`, e.g.: https://github.com/RustCrypto/KEMs/blob/master/.github/workflows/ml-kem.yml#L67

tarcieri (approved):

PR #922, opened 2025-03-08
ml-dsa: cargo build --benches in github workflow

json420:
This is something to consider after #921 lands. It makes sure the benchmarks can build against the current code.


json420:
Oh yeah, and rightfully so, my updated ml-dsa.yml is failing because #921 hasn't landed yet. So it works! :nerd_face:

json420:
Sorry about the noise of closing/reopening the PR. But it did get the tests to run again, which now pass because #921 landed.

PR #923, opened 2025-03-09
remove git dependency on hashes and hmac

baloo:


json420:
I don't know enough about the changes in hashes and hmac to saying anything too smart about this PR, but I do know that if I:

```rust
cargo add ml-dsa --git https://github.com/baloo/signatures.git --branch baloo/remove-git-dependencies
cargo add slh-dsa --git https://github.com/baloo/signatures.git --branch baloo/remove-git-dependencies
cargo build
```

Then that works! The same currently does not work from RustCrypto/signatures git master. So this is obvious forward progress for my two very most favorite signatures crates, `ml-dsa` and `slh-dsa`, haha :stuck_out_tongue:

For what it's worth, :+1: from me.

PR #928, opened 2025-03-22
slh-dsa: fix benchmark to also run verify_benches()

json420:
The sign_verify.rs benchmark was running the signing benchmarks twice. This fixes that so the verification benchmarks are also run.


baloo (approved):

json420:
I figured this is probably a good PR to add  ` - run: cargo build --benches --all-features` to the github slh-dsa.yml workflow to ensure the slh-dsa benchmarks remain in a buildable state.

I'm not 100% sure I put this line in the best place, so someone with more knowledge of this fancy slh-dsa workflow, please double check and let me know if I should make any changes. Thanks!

tarcieri (approved):

PR #932, opened 2025-04-09
ml-dsa: cargo test with --no-default-features in workflow

json420:
The CI workflow should build/test with --no-default-features to catch any issues with that configuration.

There is currently 1 warning when building ml-dsa with --no-default-features, and 1 failing doc-test. I'll fix those in upcoming commits, but I first want to run this though the CI to make sure it fails as expected.

### DIFF
--- a/.github/workflows/ml-dsa.yml
+++ b/.github/workflows/ml-dsa.yml
@@ -34,3 +34,4 @@ jobs:
       - run: cargo build --benches --all-features
       - run: cargo test --release
       - run: cargo test --all-features --release
+      - run: cargo test --no-default-features --release
--- a/.github/workflows/ml-dsa.yml
+++ b/.github/workflows/ml-dsa.yml
@@ -35,3 +35,4 @@ jobs:
       - run: cargo test --release
       - run: cargo test --all-features --release
       - run: cargo test --no-default-features --lib --tests --release
+      - run: cargo test --no-default-features --features alloc --lib --tests --release
--- a/.github/workflows/ml-dsa.yml
+++ b/.github/workflows/ml-dsa.yml
@@ -34,5 +34,5 @@ jobs:
       - run: cargo build --benches --all-features
       - run: cargo test --release
       - run: cargo test --all-features --release
-      - run: cargo test --no-default-features --lib --tests --release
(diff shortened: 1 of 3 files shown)


json420:
Worth noting that `cargo build --no-default-features --features alloc` also fails badly.

json420:
I believe this is ready for review, but I'm not certain about this change in Cargo.toml:

`alloc = ["pkcs8", "pkcs8/alloc"]`

Which is what I did to fix:

`cargo build --no-default-features --features alloc`

It gets the tests passing, but it also makes the `alloc` feature depend on the `pkcs8` feature (that's the part I'm unsure about)..

@bifurcation is it meaningful for the ml-dsa `alloc` feature to be used without the `pkcs8` feature enabled?

As far as I can tell the fix is either to do what I did above, or what I did early which was  just to remove the `?` like this:

`alloc = [ "pkcs8/alloc"]`

In which case the `alloc` feature depends on pkcs8 package with its own `alloc` feature enabled, but does not enable the `pkcs8` feature in ml-dsa. So I'm not sure which is correct.

Thanks!

baloo (approved):

bifurcation:
According to [the documentation for Cargo features](https://doc.rust-lang.org/cargo/reference/features.html):

> The `"package-name/feature-name"` syntax will also enable `package-name` if it is an optional dependency. Often this is not what you want. You can add a `?` as in `"package-name?/feature-name"` which will only enable the given feature if something else enables the optional dependency.

(As an aside: Given that, `alloc = ["pkcs8", "pkcs8/alloc"]` is redundant, since `pkcs8/alloc` will already enable `pkcs8`.)

But more importantly, it doesn't seem like the `alloc` feature should enable `pkcs8`.  In other words, it seems like the current line is correct.

The actual problem here is that there is code that requires `pkcs8` that is not marked as depending on `pkcs8`.  So the better solution here is to add those markers, so that the appropriate code is disabled.  I went ahead and implemented that:

https://github.com/json420/signatures/pull/1

json420:
@bifurcation thanks! Makes sense, much better than what I was trying to do :smile_cat:

tarcieri (approved):

PR #941, opened 2025-04-19
dsa: capture error

baloo:


ecdsa/src/der.rs
```diff
@@ -420,13 +420,22 @@ mod tests {
...
+            Signature::from_der(&[
+                der::Tag::Sequence.number().value().try_into().unwrap(),
+                0x00
+            ])
+            .is_err()
+        );
         assert!(
             Signature::from_der(&[
-                der::Tag::Sequence.into(),
+                der::Tag::Sequence.number().value().try_into().unwrap(),
```
baloo, line 436: Not exactly a fan. 

Maybe we should have a `impl TryFrom<der::Tag> for u8` or something.

baloo: and we lose the contructed.

tarcieri: There used to be a `From` impl, but I guess it was removed when tags were changed to support longer than 1-byte lengths

baloo: Yeah, and we lost the `Constructed` bit at this moment so you have to rebuild it manually. I don't know.
I think I'm fine with just hardcoding the value here with a comment.

tarcieri (approved):

PR #951, opened 2025-04-29
Automatically retry signing with random `k`

### LINKED ISSUE #950: ECDSA document signing failure cases
Currently ECDSA signing can fail in only one scenario as far AFAICS: the computed `r` or `s` are 0.
https://github.com/RustCrypto/signatures/blob/afe73a90c3848169f23da31098ac10116a3cf1e1/ecdsa/src/hazmat.rs#L124

However, [FIPS 186 mandates that if `r` or `s` are 0 a new `k` should be selected instead](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf):
> 11. If r = 0 or if s = 0, and k was generated deterministically (using the procedure in 6.3.2),
then output failure. Otherwise, if r = 0 or if s = 0, then go to Step 3.

I have no idea how other libraries handle this or if this is expected to be done by the user instead. In which case it should be documented.

daxpedda:
This PR changes all signing operations that generate a random `k` to retry with a new random `k` on failure instead of returning an error. This follows [FIPS 186](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf):
> 11. If r = 0 or if s = 0, and k was generated deterministically (using the procedure in 6.3.2),
then output failure. Otherwise, if r = 0 or if s = 0, then go to Step 3.

Resolves #950.

### DIFF
--- a/ecdsa/src/hazmat.rs
+++ b/ecdsa/src/hazmat.rs
@@ -93,6 +93,11 @@ pub fn bits2field<C: EcdsaCurve>(bits: &[u8]) -> Result<FieldBytes<C>> {
 ///
 /// ECDSA [`Signature`] and a [`RecoveryId`] which can be used to recover the verifying key for a
 /// given signature.
+///
+/// # Errors
+///
+/// This will return an error if a zero-scalar was generated. It can be tried again with a
+/// different `k`.
 #[cfg(feature = "arithmetic")]
 #[allow(non_snake_case)]
 pub fn sign_prehashed<C>(
@@ -140,6 +145,11 @@ where
 /// - `d`: signing key. MUST BE UNIFORMLY RANDOM!!!
 /// - `z`: message digest to be signed, i.e. `H(m)`. Does not have to be reduced in advance.
 /// - `ad`: optional additional data, e.g. added entropy from an RNG
+///
+/// # Errors
+///
+/// This will return an error if a zero-scalar was generated. It can be tried again with different
+/// entropy `ad`.
 ///
 /// [RFC6979]: https://datatracker.ietf.org/doc/html/rfc6979
 #[cfg(feature = "rfc6979")]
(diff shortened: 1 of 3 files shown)


tarcieri:
…but it says not to retry if RFC6979 is used, which is true everywhere

tarcieri:
I’m also not wild about this being an untestable edge case which has no probability of occurring

daxpedda:
> …but it says not to retry if RFC6979 is used, which is true everywhere except `hazmat::sign_prehashed`

Yes, but I was assuming that we are using deterministic generation of `k` just because things are very composable like that. E.g. while `RandomizedSigner` is using RFC6979, it feeds RNG data into `rfc6979::generate_k()`, which AFAICT is wrong, [the specification says](https://datatracker.ietf.org/doc/html/rfc6979#section-3.3):
> entropy_input
> Use int2octets(x) as entropy string.

Where `x` is the private key. Instead FIPS described a different algorithm to generate `k` for non-deterministic ECDSA. I don't know the history of why we are using RFC6979 to generate `k` for non-deterministic ECDSA.

I also thought I was following your suggestion here: https://github.com/RustCrypto/signatures/issues/950#issuecomment-2833604695.

> I’m also not wild about this being an untestable edge case which has no probability of occurring

I'm perfectly happy to limit this to the documentation on the `hazmat` functions alone, but I think users would appreciate to know that `RandomizedSigner` can be called in a loop if it fails instead of having to deal with an unknown failure here.

tarcieri:
ECDSA is extremely brittle when used without RFC6979: https://blog.trailofbits.com/2020/06/11/ecdsa-handle-with-care/

Less than one bit of bias in `k` can result in full secret key recovery. Using RFC6979 provides assurances that `k` will be unbiased.

Leveraging the `ad` string for additional entropy helps harden the implementation against [fault-injection attacks](https://eprint.iacr.org/2017/975.pdf).

PR #963, opened 2025-05-09
chore(deps): bump `crypto-bigint` to `0.7.0-pre.3`, use core naming conventions

baloo:
See:
 - https://github.com/RustCrypto/crypto-bigint/pull/817
Depends:
 - https://github.com/RustCrypto/traits/pull/1845

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,9 +267,9 @@ checksum = "7a81dae078cea95a014a339291cec439d2f232ebe854a9d672b796c6afafa9b7"
 
 [[package]]
 name = "crypto-bigint"
-version = "0.7.0-pre.2"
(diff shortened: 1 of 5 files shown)


tarcieri:
You can bump to `pre.3` now. It would be good to do that on traits as well

baloo:
Is there a reason not to go with `rc.0`? 
https://github.com/RustCrypto/crypto-bigint/issues/752 We have all the breaking changes in I think?

tarcieri:
We can call the next release `rc.0` I guess, though there's still the question of switching to binary (X)GCD which would involve some breaking API changes

baloo:
Alright, let’s wait for gcd then

tarcieri (approved):

PR #965, opened 2025-05-12
ecdsa: der: use impl `EncodeValue`/`DecodeValue` on `SignatureRef`

dishmaker:
This way we avoid using internals of `der`:
- `der::SliceWriter`
- `der::SliceReader`
- `tag.assert_eq`
- `writer.sequence`

### DIFF
--- a/ecdsa/src/der.rs
+++ b/ecdsa/src/der.rs
@@ -385,11 +385,6 @@ impl<'a> DecodeValue<'a> for SignatureRef<'a> {
 }
 impl<'a> Sequence<'a> for SignatureRef<'a> {}
 
-/// Decode the `r` and `s` components of a DER-encoded ECDSA signature.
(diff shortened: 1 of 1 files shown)


ecdsa/src/der.rs
```diff
@@ -82,7 +85,7 @@ where
 {
     /// Parse signature from DER-encoded bytes.
     pub fn from_bytes(input: &[u8]) -> Result<Self> {
-        let (r, s) = decode_der(input).map_err(|_| Error::new())?;
+        let SignatureRef { r, s } = decode_der(input).map_err(|_| Error::new())?;
```
tarcieri, line 88:
```suggestion
        let SignatureRef { r, s } = SignatureRef::from_der(input).map_err(|_| Error::new())?;
```

PR #969, opened 2025-05-13
dsa: reduce `Box<[u8]>` allocation in `SignatureEncoding::to_vec`

dishmaker:
Rewritten as `SignatureRef`

### DIFF
--- a/dsa/src/lib.rs
+++ b/dsa/src/lib.rs
@@ -109,43 +109,99 @@ impl Signature {
     pub fn s(&self) -> &NonZero<BoxedUint> {
         &self.s
     }
+
+    fn to_boxed(&self) -> SignatureBoxed {
+        SignatureBoxed {
+            r: self.r.to_be_bytes(),
+            s: self.s.to_be_bytes(),
+        }
+    }
+    fn to_der_using_ref(&self) -> der::Result<Vec<u8>> {
+        self.to_boxed().to_ref()?.to_der()
+    }
 }
+
+struct SignatureBoxed {
+    r: Box<[u8]>,
+    s: Box<[u8]>,
+}
+impl SignatureBoxed {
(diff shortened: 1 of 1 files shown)


dsa/src/signature_ref.rs
```diff
@@ -0,0 +1,78 @@
+use alloc::boxed::Box;
+use crypto_bigint::{BoxedUint, NonZero};
+use pkcs8::der::{
+    self, Decode, DecodeValue, Encode, EncodeValue, FixedTag, Header, Length, Reader, Sequence,
+    Writer, asn1::UintRef,
+};
+
+use crate::Signature;
+
+pub(crate) struct SignatureBoxed {
```
baloo, line 10: That is essentially what `dsa::Signature` is (dsa::Signature is composed of `NonZero<BoxedUint>`)

baloo: wouldn't it be possible to get an iterator over the inner limbs? and get big endian from there? without allocating a new box.

dishmaker: `der` checks for leading zeros when encoding UIntRef, so it would be too tricky

baloo: well that's just up to the consumer of the said iterator to strip leading zeroes?

baloo: yeah nevermind, you have to have a buffer for `SignatureRef`, that can't be an iterator of limbs.

tarcieri:
The `crypto-bigint` crate has native support for `der`, although it's missing `DecodeValue`/`EncodeValue` impls for `BoxedUint` (and `NonZero<BoxedUint>`). They exist for `Uint`.

Perhaps support could be added there so there's not need for complex encoding logic in the `dsa` crate.

tarcieri:
Closing as stale

PR #982, opened 2025-06-01
Implement `MultiPartSigner/Verifier` for ML-DSA and SLH-DSA

daxpedda:
Implementation of `MultiPartSigner` and `MultiPartVerifier` added in https://github.com/RustCrypto/traits/pull/1880.

I only added it for ML-DSA and SLH-DSA. LMS could potentially make use of it as well.
Notably the traits aren't added for ECDSA. I felt it was kinda wrong seeing that users *should* use pre-hashing instead.
Let me know if you want me to add it somewhere else as well.

The implementation in ML-DSA and SLH-DSA is not ideal. Basically all the relevant functions are public and only take `&[u8]`. So instead of changing public functions, I just moved the implementation of those functions into private functions that take `&[&[u8]]`. Let me know if you want me to solve this in any other way.


tarcieri:
It would be good to add it to `ecdsa` for completeness, so the trait works over all signature algorithms.

(Maybe we can add a macro to write impls for types which impl `DigestSigner`, or possibly a blanket impl, but start with `ecdsa` first)

daxpedda:
Done.

Just leaving a list of potentially other applicable implementations:
- `bign256` (not covered by `ecdsa`)
- `sm2` (not covered by `ecdsa`)
- DSA
- LMS

tarcieri:
There's also `rsa` (both RSASSA-PKCS#1v1.5 and PSS)

daxpedda:
> There's also `rsa` (both RSASSA-PKCS#1v1.5 and PSS)

Let me know if you want me to add if there as well.
Will add an up-to-date list in https://github.com/RustCrypto/traits/pull/1880.

ml-dsa/src/lib.rs
```diff
@@ -168,10 +168,10 @@ where
 // This method takes a slice of slices so that we can accommodate the varying calculations (direct
 // for test vectors, 0... for sign/sign_deterministic, 1... for the pre-hashed version) without
 // having to allocate memory for components.
-fn message_representative(tr: &[u8], Mp: &[&[u8]]) -> B64 {
+fn message_representative(tr: &[u8], Mp: &[&[&[u8]]]) -> B64 {
     let mut h = H::default().absorb(tr);
 
-    for m in Mp {
+    for m in Mp.iter().copied().flatten() {
```
tarcieri, line 174: Hmm, is the `&[&[&[u8]]]` really necessary here?

PR #983, opened 2025-06-03
ed448: make ComponentBytes an array

baloo:


ed448/src/lib.rs.orig
```diff
@@ -0,0 +1,201 @@
+#![no_std]
```
tarcieri, line 1: Think you forgot to remove this

baloo: sorry, bitten by jujutsu's autosnapshot feature.

PR #990, opened 2025-06-05
ed448: drop `-signature` suffix from crate name

tarcieri:
Crate name is now just `ed448`, like the Ed448 variant of EdDSA.

The elliptic curve's name is Ed448-Goldilocks (or "edwards448")


kevaundray:
> @kevaundray can you add me as an owner of the `ed448` crate?

I just did it, please confirm it went through :)

tarcieri:
@kevaundray it did, thanks!

PR #1008, opened 2025-06-30
feat(ml-dsa): Implement verifying_key method for SigningKey

romen:
Introduce a utility method to derive a `VerifyingKey` from a `SigningKey`, even when the original seed or precomputed public key is unavailable. This aids in scenarios where private keys are imported from external sources lacking associated public key data.

### DIFF
--- a/ml-dsa/src/lib.rs
+++ b/ml-dsa/src/lib.rs
@@ -504,6 +504,23 @@ impl<P: MlDsaParams> SigningKey<P> {
             None,
         )
     }
+
+    /// This auxiliary function derives a `VerifyingKey` from a bare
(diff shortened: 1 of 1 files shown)


tarcieri:
It would be good to add an impl of the [`signature::Keypair`] trait: https://docs.rs/signature/3.0.0-rc.1/signature/trait.Keypair.html

You can keep the inherent method too so it still works with the trait out-of-scope

romen:
@tarcieri sorry, it took some time.

I rebased this PR on latest master, and added the trait implementation.

The inherent method is now just a convenience shim around calling the trait implementation.

PR #1030, opened 2025-08-05
ecdsa: use new `Reduce` trait

tarcieri:
Uses the `Reduce` trait added in RustCrypto/traits#1949


tarcieri:
@baloo the updates to https://github.com/RustCrypto/elliptic-curves look a bit involved. If you want to give it a try that's cool, otherwise I'll take a look at it tomorrow

Issue #1045, opened 2025-09-01
ml-dsa: feature request for a `SigningKey::from_seed(B32)` method


hoxxep:
TLDR: feature request for `SigningKey::from_seed(B32)` in ml-dsa.

Keys are currently generated using `MlDsa65::key_gen(&mut rng) -> KeyPair<MlDsa65>`, and then the signing key may be stored via `SigningKey::encode(&self)` which produces a 4032 byte vector. 

I can see that `key_gen` generates a 32 byte seed vector internally, and then constructs the signing key from this seed. To make storing the signing keys more efficient, would it be possible to include a `SigningKey::<MlDsa65>::from_seed(B32)` method to allow a user to construct a signing key from a 32-byte vector directly, so that a signing user only needs to store the seed?

I didn't want to start using `key_gen_internal` as there's a TODO intending for it to be marked as a private method. The proposed method would also not construct the verifying key from the seed, and avoid this work until `SigningKey::verifying_key(&self)` is called.

Cheers!

PR #1046, opened 2025-09-02
ml-dsa: SigningKey::from_seed(&B32) implementation

hoxxep:
Providing an implementation for `SigningKey::from_seed` as suggested in issue #1045. 

The `from_seed` implementation copies code from `KeyGen::key_gen_internal`, `VerifyingKey::new`, and `VerifyingKey::encode_internal`, omitting a small number of lines that were verifying-key specific work.

The tests are a simple comparison to `key_gen_internal` which also may not be ideal.

Happy to replace this PR for any of the concerns above, or if there's a better alternative API that's perhaps more consistent with the other crates. Thanks!


tarcieri:
> The from_seed implementation copies code from KeyGen::key_gen_internal, VerifyingKey::new, and VerifyingKey::encode_internal, omitting a small number of lines that were verifying-key specific work.

Ideally the duplication could be eliminated

hoxxep:
Agreed given the method is duplicating 25 lines of complicated maths. It is only duplicated once though.

This is `VerifyingKey::new`'s existing type signature and implementation, commenting the logic that's necessary to compute a `SigningKey`.

```rust
    fn new(
        rho: B32,
        t1: Vector<P::K>,
        A_hat: Option<NttMatrix<P::K, P::L>>,
        enc: Option<EncodedVerifyingKey<P>>,
    ) -> Self {
        // omitted in SigningKey::from_seed
        let A_hat = A_hat.unwrap_or_else(|| expand_a(&rho));
        let t1_2d_hat = (Elem::new(1 << 13) * &t1).ntt();

        // used for SigningKey::from_seed, with `enc == None`
        let enc = enc.unwrap_or_else(|| Self::encode_internal(&rho, &t1));
        let tr: B64 = H::default().absorb(&enc).squeeze_new();

        Self { ... }
    }
```

This method gets called by `VerifyingKey::decode`, `KeyGen::key_gen_internal`, and `SigningKey::verifying_key`. I can't immediately think of a good way to untangle it that doesn't cause extra computations.

Will also get those clippy warnings fixed.

hoxxep:
I've benchmarked this PR's `SigningKey::from_seed` implementation as only being 10% faster by omitting the couple of extra steps to also compute a `VerifyingKey`.

I like that the other implementation (#1054) doesn't duplicate the logic, and I'm assuming a 20us or 10% performance difference in key derivation is negligible that we can always optimise later in another PR if required.

Benchmarking results of this PR below:
```
keygen/keypair          time:   [217.63 µs 218.17 µs 218.77 µs]
keygen/signingkey       time:   [196.36 µs 196.99 µs 197.68 µs]
```

Benchmark implementations, for reference:
```rust
    c.bench_function("keygen/keypair", |b| {
        b.iter(|| {
            let kp = MlDsa65::key_gen_internal(black_box(&xi));
            let sk_bytes = kp.signing_key().encode();
            black_box(sk_bytes)
        })
    });

    c.bench_function("keygen/signingkey", |b| {
        b.iter(|| {
            let sk_bytes = SigningKey::<MlDsa65>::from_seed(black_box(&xi)).encode();
            black_box(sk_bytes)
        })
    });
```

I will close this PR in favour of #1054 given the negligible performance difference. I'm happy to re-implement this PR at a later date if desired.

PR #1054, opened 2025-09-03
ml-dsa: Rename `KeyGen::key_gen_internal` to `from_seed`

hoxxep:
This is an alternative to #1046 to reduce copying the seed derivation logic. Implementing a `from_seed` as raised in issue #1045.

- Renames `KeyGen::key_gen_internal` to `KeyGen::from_seed`.
- Adds `SigningKey::from_seed`. This internally calls `KeyGen::from_seed` for now, as the majority of the computational work is the same for both keys. A future PR could be slightly more efficient by specialising the SigningKey from_seed method by skipping the couple of extra steps performed to also compute a VerifyingKey, but #1046 shows this only nets a 20us / 10% speed up.
- Adds a test to the `from_seed` implementations don't deviate from each other in future, if more efficient implementations are provided.

I am not sure why `key_gen_internal` was commented to be behind a feature flag or private, as I believe deriving from a seed value is a must-have, assuming this implementation correctly matches the spec. I can look through the spec and codebase for test vectors to make sure, if desired?

Cheers!

### DIFF
--- a/ml-dsa/src/lib.rs
+++ b/ml-dsa/src/lib.rs
@@ -237,7 +237,7 @@ where
 
         let seed = Array::try_from(private_key_info.private_key.as_bytes())
             .map_err(|_| pkcs8::Error::KeyMalformed)?;
-        Ok(P::key_gen_internal(&seed))
+        Ok(P::from_seed(&seed))
     }
 }
 
@@ -350,6 +350,16 @@ impl<P: MlDsaParams> SigningKey<P> {
             t0_hat,
             A_hat,
         }
+    }
+
+    /// Deterministically generate a signing key from the specified seed.
+    ///
+    /// This method reflects the ML-DSA.KeyGen_internal algorithm from FIPS 204, but only returns a
+    /// signing key.
+    #[must_use]
+    pub fn from_seed(seed: &B32) -> Self {
+        let kp = P::from_seed(seed);
+        kp.signing_key
     }
 
     /// This method reflects the ML-DSA.Sign_internal algorithm from FIPS 204. It does not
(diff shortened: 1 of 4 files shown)


ml-dsa/src/lib.rs
```diff
@@ -913,8 +927,7 @@ pub trait KeyGen: MlDsaParams {
     fn key_gen<R: CryptoRng + ?Sized>(rng: &mut R) -> Self::KeyPair;
 
     /// Deterministically generate a signing key pair from the specified seed
-    // TODO(RLB): Only expose this based on a feature.
```
hoxxep, line 916: I am not sure of this comment in particular. There may be reasons `key_gen_internal` cannot be named `from_seed` that I'm not aware of, and I can't think of reasons for feature gating this? cc. @tarcieri @bifurcation

tarcieri: The name comes from [FIPS 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf), namely section 6.1 where it is referred to as `ML-DSA.KeyGen_internal`.

I don't oppose the change in this PR, though, just providing a reference for why it's named the way it is.

hoxxep: Thanks, I've updated the docstrings to include that for end users.

ml-dsa/src/lib.rs
```diff
@@ -668,6 +678,16 @@ impl<P: MlDsaParams> VerifyingKey<P> {
...
 
+    /// Deterministically generate a verifying key from the specified seed.
+    ///
+    /// This method reflects the ML-DSA.KeyGen_internal algorithm from FIPS 204, but only returns a
+    /// verifying key.
+    #[must_use]
+    pub fn from_seed(seed: &B32) -> Self {
+        let kp = P::from_seed(seed);
+        kp.verifying_key
+    }
```
tarcieri, line 689: I'm a little bit wary about an API like this on `VerifyingKey` because it might lead a user to confuse their seed with a public key.

How about putting `from_seed` on `KeyPair`, and then the user can call `verifying_key()` instead? It also avoids the duplication of having two `from_seed` methods.

PR #1059, opened 2025-09-04
dsa: add `from_components_unchecked` APIs

### LINKED ISSUE #1058: dsa: support for different size of keys
Context: PGP has some very old keys including some very weak DSA keys (768 / 960 bit keys). But also some odd-ball 4096 bit keys (that I can't find a documentation/standard for). See https://github.com/rpgp/rpgp/pull/573#issuecomment-3249986230

Since we migrated dsa to crypto-bigint (#906), dsa will only accept key sizes specified by NIST:
https://github.com/RustCrypto/signatures/blob/51dd7d8a71e5e5bedd8f52258121c75f19a2e55c/dsa/src/size.rs#L18-L27

NIST standard NIST.FIPS.186-4 Section `4.2 Selection of Parameter Sizes and Hash Functions for DSA`:
https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf#page=25

 - Is there a will to support older / weak key sizes (under an hazmat flag)?

 - Should we just skip the key size checks? A `Components::from_components_unchecked` method?
https://github.com/RustCrypto/signatures/blob/51dd7d8a71e5e5bedd8f52258121c75f19a2e55c/dsa/src/components.rs#L48-L55

cc @hko-s @dignifiedquire

baloo:
This intends to add support for any size of keys, either larger or older/weak keys.

The API is only available under the `hazmat` feature-flag, and the consumer is responsible for checking the security of what they're using it for.

Fixes #1058

cc @dignifiedquire @hko-s

### DIFF
--- a/dsa/src/size.rs
+++ b/dsa/src/size.rs
@@ -1,7 +1,8 @@
+use core::cmp::Ordering;
 use crypto_bigint::Limb;
 
 /// DSA key size
-#[derive(Clone, Debug, Copy, PartialEq, PartialOrd)]
+#[derive(Clone, Debug, Copy)]
 pub struct KeySize {
     /// Bit size of p
     pub(crate) l: u32,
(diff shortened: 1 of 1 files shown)


dsa/src/components.rs
```diff
@@ -31,6 +31,44 @@ pub struct Components {
...
+    ///
+    /// Any length of keys may be used, no checks are to be performed. You are responsible for
+    /// checking the key strengths.
+    #[cfg(feature = "hazmat")]
+    pub fn from_components_unchecked(
+        p: BoxedUint,
+        q: BoxedUint,
+        g: BoxedUint,
+    ) -> signature::Result<Self> {
+        let (p, q, g) = Self::adapt_components(p, q, g)?;
```
tarcieri, line 60: Instead of `adapt_components`, could `from_components` call `from_components_unchecked` which contains `adapt_components` inlined, and then apply its key size checks?

tarcieri: Aah okay, makes sense

PR #1062, opened 2025-09-10
chore: apply `unreachable_pub` lint

TomAFrench:
Continues on with updates started in https://github.com/RustCrypto/hashes/pull/734. Still needs a couple of crates to be dealt with.

### DIFF
--- a/ecdsa/src/dev.rs
+++ b/ecdsa/src/dev.rs
@@ -224,7 +224,7 @@ macro_rules! new_wycheproof_test {
             }
 
             impl TestVector {
-                pub fn pass(&self) -> bool {
+                pub(crate) fn pass(&self) -> bool {
                     match self.pass_ {
                         &[0] => false,
                         &[1] => true,


ml-dsa/src/param.rs
```diff
@@ -31,20 +31,21 @@ use crate::encode::{
...
-pub type Pow2DMinus1Minus1 = Diff<Pow2DMinus1, U1>;
+pub(crate) type SpecQ = Sum<Diff<Shleft<U1, U23>, Shleft<U1, U13>>, U1>;
+pub(crate) type SpecD = U13;
+pub(crate) type QMinus1 = Diff<SpecQ, U1>;
+pub(crate) type BitlenQMinusD = Diff<Length<SpecQ>, SpecD>;
+pub(crate) type Pow2DMinus1 = Shleft<U1, Diff<SpecD, U1>>;
+pub(crate) type Pow2DMinus1Minus1 = Diff<Pow2DMinus1, U1>;
 
 /// An integer that describes a bit length to be used in sampling
+#[expect(unreachable_pub)]
```
TomAFrench, line 42: There's a couple of instances of unreachable pub traits which need to remain pub due to trait constraints on associated types in pub traits. I've silenced the warning without making these reachable in these cases.

TomAFrench:
huh, this was passing locally. Will take another look.

PR #1075, opened 2025-09-15
`slh-dsa`: implement `DigestSigner/Verifier`

daxpedda:
Follow-up to https://github.com/RustCrypto/elliptic-curves/pull/1423.


slh-dsa/src/hashes/sha2.rs
```diff
@@ -117,7 +144,8 @@ where
             .chain_update(pk_seed)
             .chain_update(&zeroes)
             .chain_update(adrs.compressed());
-        m.iter().for_each(|x| sha.update(x.as_slice()));
+        m.iter()
+            .for_each(|x| Update::update(&mut sha, x.as_slice()));
```
daxpedda, line 148: Hitting the same problem as usual. This time really unclear how to avoid it without loosing access to all the QoL methods of `Digest`.

slh-dsa/src/hashes.rs
```diff
@@ -7,32 +7,43 @@ mod shake;
...
 use crate::{PkSeed, SkPrf, SkSeed, address::Address};
 
+/// `Digest` parameter for [`DigestSigner`] and [`DigestVerifier`].
+///
+/// [`DigestSigner`]: signature::DigestSigner
+/// [`DigestVerifier`]: signature::DigestVerifier
+pub trait HashDigest {
+    /// The associated `Digest` type.
+    type Digest: Update;
+}
```
daxpedda, line 26: I originally had `type Digest` on `HashSuite` directly, but that made it publicly not accessible.

So I introduced `HashDigest` just to make it publicly accessible if users ever need to access the type.

PR #1082, opened 2025-10-08
ecdsa: re-export blobby

baloo:
Re-exporting blobby directly from ecdsa makes version upgrades a tiny bit more resilient to API changes made in blobby.

They otherwise have to be carefully managed between ecdsa and its dependencies.

If the binary format changes between the blobby upgrade, the blob files will still need to be updated to match.

### DIFF
--- a/ecdsa/src/dev.rs
+++ b/ecdsa/src/dev.rs
@@ -150,7 +150,7 @@ macro_rules! new_wycheproof_test {
     ($name:ident, $test_name: expr, $curve:path) => {
         use $crate::{
             Signature,
-            elliptic_curve::{sec1::EncodedPoint},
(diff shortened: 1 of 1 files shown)


ecdsa/src/dev.rs
```diff
@@ -148,14 +150,13 @@ macro_rules! new_wycheproof_test {
     ($name:ident, $test_name: expr, $curve:path) => {
         use $crate::{
             Signature,
-            elliptic_curve::{bigint::Integer, sec1::EncodedPoint},
+            elliptic_curve::{sec1::EncodedPoint},
```
tarcieri:
```suggestion
            elliptic_curve::sec1::EncodedPoint,
```

PR #1305, opened 2026-04-16
Introduce API to parse PEM keys in a simpler way

DoumanAsh:
As per idea from https://github.com/RustCrypto/elliptic-curves/issues/1733#issuecomment-4257049505

There is too many micro dependencies between each other but...
I cooked up `SigningKey::from_pem` method that parses PEM content depending on label as simple as possible
At the end of the day it requires `pem`, `der` and optionally `pkcs8` with `alloc`
I initially thought `sec1` lives on its own, but right now it seems to be Interleaved with `pkcs8` so I could make it required feature instead since there is not much benefit in having code path for builds that don't have `pkcs8` with `alloc`


tarcieri:
This should probably go in `elliptic-curve` and be defined on `SecretKey`, then `ecdsa` can delegate to that

DoumanAsh:
Created PR 
https://github.com/RustCrypto/traits/pull/2387

I see that it prefers to hide error meaning, but for now I think this case would benefit from having proper error to indicate what exactly is wrong, but let me know if I need to hide error anyway

tarcieri:
@DoumanAsh can you update this PR to delegate to the API added to `elliptic-curve`?

DoumanAsh:
Thanks for reminder, I will rebase PR and update it to use new method

DoumanAsh:
@tarcieri I rebased and implemented `from_pem` method using new API
I hid error from public API via generic for now but didn't add any tests since it just delegates to underlying API of SecretKey

PR #1308, opened 2026-04-26
ml-dsa: zeroize seed on SigningKey drop

MavenRain:
The `SigningKey` (formerly `KeyPair`) held the 32-byte `seed` directly,
  and `Array<u8, U32>` does not pick up `ZeroizeOnDrop` because `u8`
  does not impl `ZeroizeOnDrop`. The seed is private key material, so it
  needs an explicit `Drop` that zeroizes.

  Adding `Drop` to `SigningKey` broke two call sites that moved
  `signing_key` out of a `SigningKey` (Rust forbids field moves out of
  `Drop` types). Both are refactored: `ExpandedSigningKey::from_seed`
  now contains the FIPS-204 KeyGen_internal logic directly, and
  `P::from_seed` delegates to it. The pkcs8 `TryFrom<PrivateKeyInfoRef>`
  for `ExpandedSigningKey` decodes the seed inline and calls
  `ExpandedSigningKey::from_seed` rather than going through `SigningKey`.

  Adds a compile-time test asserting `ZeroizeOnDrop` on both
  `SigningKey<P>` and `ExpandedSigningKey<P>` for each parameter set.

### DIFF
--- a/ml-dsa/src/lib.rs
+++ b/ml-dsa/src/lib.rs
@@ -269,6 +269,18 @@ impl<P: MlDsaParams> CtEq for SigningKey<P> {
     }
 }
 
+#[cfg(feature = "zeroize")]
+impl<P: MlDsaParams> Drop for SigningKey<P> {
+    fn drop(&mut self) {
+        // `signing_key` is zeroized by its own `Drop` impl. Zeroize the seed,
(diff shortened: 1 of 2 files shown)


tarcieri:
Thanks for pointing this out. I'm taking care of it in #1345.

PR #1310, opened 2026-04-27
ecdsa: use absolute 128-bit floor for bits2field

MavenRain:
Per maintainer suggestion on #1275, replace the curve-specific
  `FieldBytesSize / 2` minimum with an absolute 16-byte / 128-bit
  floor. The old rule rejected legitimate pairings like SHA-256 with
  P-521 (permitted by e.g. XML Signature); the new rule still catches
  egregious digest misuse without those false positives.

  Closes #1275


tarcieri (approved):
We might consider raising this above 128-bits in the future (224-bits might be considered the absolute minimum, IMO) but this addresses the immediate issue

PR #1311, opened 2026-04-27
ml-dsa: fix `unreachable_pub` lint by making `Eta` pub

tarcieri:
This lint started failing, seemingly as of Rust 1.95. Though this may not be the best solution, to get the lint to pass again this makes `Eta` pub.


tarcieri:
Ugh, what passes `unreachable_pub` on 1.95.0 does not pass on 1.85.0. I just added `allow(unreachable_pub)` on the offending items for now.

PR #1360, opened 2026-05-22
rfc6979: allow the use of non-block_api-backed hashes

baloo:
When https://github.com/rustcrypto/signatures/pull/1076 was introduced, it moved the implementation to use block_api-backed hashes only.

This made sense at the time, because all the hashes used so far were backed by block_api.

This situation changed with https://github.com/RustCrypto/hashes/pull/849 where sha3 is no longer backed by the block_api.

This PR is essentially a revert of https://github.com/RustCrypto/hashes/pull/849 that brings back the previous implementation of rfc6979, using SimpleHmacReset.

This was tested in downstream crates (k256 and rpgp).

See discussion in https://github.com/RustCrypto/signatures/pull/1076#issuecomment-4501887654 and below

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1139,6 +1139,7 @@ dependencies = [
  "hex-literal",
  "hmac",
  "sha2",
+ "sha3 0.12.0",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


baloo:
cc @daxpedda

daxpedda:
I believe this was done mainly to avoid `Digest + Update`, which you aren't doing here. I don't remember exactly why this was necessary in the first place (https://github.com/RustCrypto/elliptic-curves/pull/1423 was the trigger AFAICS).

Ergo this looks fine to me :+1:. Thank you for the heads-up!

tarcieri:
Can you also bump the version of `rfc6979` to v0.6.0-pre since this is a breaking change? (unless it isn't?)

PR #1369, opened 2026-06-05
dsa: add Wycheproof verification test vectors

### LINKED ISSUE #816: dsa: Add the Wycheproof test vectors to the test suite
The test suite currently only tests against a few RFC test vectors, importing the Wycheproof vectors would be nice to improve the confidence in the correctness of the implementation.

arpitjain099:
Closes #816

The `dsa` test suite previously only covered a handful of RFC vectors. This adds a `dsa/tests/wycheproof.rs` suite that runs the C2SP Wycheproof DSA test vectors through the crate's verification path, which gives much broader coverage of the malformed-input and edge-case space.

## What it does

The vectors come from the `thirdparty/wycheproof` submodule that is already wired into this repo (the same source the `ml-dsa` Wycheproof tests use), so there is no new vendored data and no new runtime dependency. Run `git submodule update --init` before running the suite.

Four test groups are exercised, 1432 vectors in total:

- `dsa_2048_224_sha224_test.json` (336 vectors)
- `dsa_2048_224_sha256_test.json` (364 vectors)
- `dsa_2048_256_sha256_test.json` (366 vectors)
- `dsa_3072_256_sha256_test.json` (366 vectors)

For each group the test loads the DER `SubjectPublicKeyInfo` into a `VerifyingKey`, parses the DER-encoded signature into a `Signature`, and verifies it with the group's hash (SHA-224 or SHA-256) through `DigestVerifier`. The observed accept/reject result is asserted against the Wycheproof expected result. Valid vectors that parse are additionally cross-checked through the `PrehashVerifier` entrypoint so both verification APIs stay in agreement.

Breakdown across the four files: 296 `valid`, 1132 `invalid`, 4 `acceptable`.

## Edge cases handled explicitly

- **`acceptable` / `MissingZero`**: the only `acceptable`-class vectors in these files carry the `MissingZero` flag, a legacy ASN.1 integer for `r` that omits its leading `0x00` padding byte. The crate requires canonical DER integers (`UintRef`) and rejects these, which is conformant because Wycheproof allows an implementation to either accept or reject an `acceptable` vector. This is handled with a comment rather than skipped silently.
- **Malformed signature encodings**: a signature that fails to parse as canonical DER `SEQUENCE { r, s }` is treated as a verification failure. That is the desired behaviour for the `invalid` cases and would only be a regression for a `valid` case (which the assertion catches).
- **P1363 files not loaded**: the sibling `*_p1363_test.json` files use the raw `r || s` IEEE P1363 signature encoding. The `dsa` crate parses only DER signatures, so those files are intentionally not loaded; this is documented at the top of the test file.

## Testing

```
cargo test -p dsa --test wycheproof
```

```
running 4 tests
test dsa_2048_224_sha224 ... ok
test dsa_2048_224_sha256 ... ok
test dsa_2048_256_sha256 ... ok
test dsa_3072_256_sha256 ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

The full `cargo test -p dsa` suite and `cargo clippy -p dsa --tests --all-features` are both green.

Two small dev-dependencies were added to support the JSON loading (`serde` and `serde_json`, matching the versions already used by the `ml-dsa` Wycheproof tests), and the `serde` feature was enabled on the existing `hex` dev-dependency.

### DIFF
--- a/.github/workflows/dsa.yml
+++ b/.github/workflows/dsa.yml
@@ -58,6 +58,8 @@ jobs:
           git config --global core.eol lf
 
       - uses: actions/checkout@v6.0.2
+        with:
+          submodules: recursive
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.toolchain }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -449,6 +449,8 @@ dependencies = [
  "proptest",
  "rand_core 0.10.1",
  "rfc6979",
+ "serde",
+ "serde_json",
  "sha1",
  "sha2",
  "signature",
--- a/dsa/Cargo.toml
+++ b/dsa/Cargo.toml
@@ -31,12 +31,14 @@ pkcs8 = { version = "0.11", optional = true, default-features = false, features 
 [dev-dependencies]
 chacha20 = { version = "0.10", features = ["rng"] }
 der = { version = "0.8", features = ["derive"] }
-hex = "0.4"
+hex = { version = "0.4", features = ["serde"] }
 hex-literal = "1"
 getrandom = { version = "0.4", features = ["sys_rng"] }
 pkcs8 = { version = "0.11", default-features = false, features = ["pem"] }
 proptest = "1"
 rand_core = "0.10"
+serde = { version = "1.0.215", features = ["derive"] }
+serde_json = "1.0.132"
 sha1 = "0.11"
 
 [features]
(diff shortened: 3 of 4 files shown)


arpitjain099:
The all-platform failure was just CI not checking out the `thirdparty/wycheproof` submodule, so the vector files were missing. Added `submodules: recursive` to the dsa test job's checkout (matching the ml-dsa workflow). With the submodule present the 4 DSA wycheproof tests pass locally.

PR #1370, opened 2026-06-05
ml-dsa: activate `MaybeBox` with alloc feature

baloo:
When the matrices were pushed on the stack in #1344 and #1345, it relies on the `MaybeBox` implementation of `module-lattice`.

`MaybeBox` will push the matrix on the heap only if `alloc` is enabled.

Because `alloc` never got activated in `ml-dsa` this effectively reverted the optimization on the stack size made in #1261 and #1259.

This commit fixes that by enabling `alloc` on `module-lattice` when `alloc` is enabled on `ml-dsa` and restoring the expected behavior of `MaybeBox`.

### DIFF
--- a/ml-dsa/Cargo.toml
+++ b/ml-dsa/Cargo.toml
@@ -26,7 +26,7 @@ exclude = [
 
 [features]
 default = ["alloc", "getrandom", "pkcs8"]
-alloc = ["hybrid-array/alloc", "pkcs8?/alloc", "signature/alloc"]
+alloc = ["hybrid-array/alloc", "module-lattice/alloc", "pkcs8?/alloc", "signature/alloc"]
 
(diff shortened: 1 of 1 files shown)


baloo:
a workaround for a consumer crate until this makes into a release is to enable the feature on `module-lattice` manually:
```
[dependencies]
ml-dsa = { version = "0.1", features = ["alloc"] }
module-lattice = { version = "0.2.3", features = ["alloc"] }
```

tarcieri:
This looks like a dup of #1365? I think if you rebase on `master` it won't have any changes.

baloo:
indeed, sorry for the noise.

tarcieri:
I can cut a patch release with the fix

baloo:
I think we can use the workaround, if you have time that'd be appreciated, but it's fine otherwise.

PR #1386, opened 2026-06-17
ml-dsa: fix constant time issues

lukaszobernig:
I saw that these two spots still used non-constant time code, let's replace this with branchless implementations instead.

* We only call `coeff_from_half_byte` during key generation, which means it is a lot harder for an attacker to collect enough execution traces. But since we generally use private key seeds, it's usually not just a one-shot for a fixed input and there is a chance to collect more traces. So it's definitely worth fixing in my opinion.

* The change to `Vector::infinity_norm` is not absolutely required, since by https://pq-crystals.org/dilithium/data/dilithium-specification-round3.pdf, section 5.5, it is OK to leak which coefficient violates the bound as long as we do not leak the coefficient itself or its sign. The change to `Polynomial::infinity_norm` although is important since it fixes that potential leak of information about the individual coefficients. Let me know if you would want to keep `Vector::infinity_norm` as is, this would probably save a tiny amount of instruction cycles.

### DIFF
--- a/ml-dsa/src/algebra.rs
+++ b/ml-dsa/src/algebra.rs
@@ -186,8 +186,7 @@ impl AlgebraExt for Polynomial {
         self.0
             .iter()
             .map(AlgebraExt::infinity_norm)
-            .max()
-            .expect("should have a maximum")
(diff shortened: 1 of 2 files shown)


tarcieri (approved):
Thanks!

PR #1395, opened 2026-06-28
rfc6979: integrate with `crypto-bigint`; `KGenerator` API

tarcieri:
Implements functions which are a 1:1 mapping to the ones defined in RFC6979, operating over `U: Unsigned + Encoding`.

Notably this also implements the modular reduction instead of leaving it to the caller to perform. As the RFC notes, it's easily implemented as conditional subtraction.

The public API has been changed to a `KGenerator` struct which provides an iterator-like API that can be called repeatedly if needed, e.g. if the computed `r` is zero.

This makes it possible to make the `ecdsa::hazmat` RFC6979-based functions infallible, because they can automatically retry with the next `k` until they find one that works.


tarcieri:
@baloo this is basically working, but `dsa` and `ecdsa` need to be updated

tarcieri:
Alright, `ecdsa` is working! 🎉 

Now for the really fun part: `dsa`

baloo:
~~I don't believe there is a way to make it work for `dsa` (Q being dynamically sized). I think it's worth duplicating the old rfc6979 implementation in a submodule of dsa and calling it a day.~~

EDIT: nevermind, BoxedUint implements Unsigned and Encoding.

https://github.com/RustCrypto/signatures/pull/1396

baloo (approved):
Tested in rpgp (https://github.com/rpgp/rpgp/pull/573 (with the changes in my PR))
They have a fair amount of edge-cases for dsa over there.

tarcieri:
@baloo it's designed to be generic so you can use it with `BoxedUint` too (actually just added a bunch of tests for that).

I'm currently debugging a `dsa` test failure, but I think I'm close. Getting confused by the hash truncation. I can push up a WIP.

Edit: hah, guess you got it working!

PR #1405, opened 2026-06-30
dsa: make `secret_number_rfc6979` infallible; cleanups

tarcieri:
The previous function was doing quite a bit, so this decomposes it into a few smaller functions: `init_kgen`, `bytes2uint`, and `qlen`.


tarcieri:
@baloo can you double check this? if it looks good, I think I can cut a final release after merging

baloo:
all the tests of rpgp are passing.

Otherwise changes look good.

PR #1419, opened 2026-07-08
ml-dsa: push the heavy bits of the signature on the heap

baloo:
Before this commit, the `ml_dsa::Signature` would cost up to 2120 bytes on the stack.

This pushes the Hint to the heap if alloc feature is enabled.

### DIFF
--- a/ml-dsa/src/hint.rs
+++ b/ml-dsa/src/hint.rs
@@ -7,7 +7,7 @@ use hybrid_array::{
     Array,
     typenum::{U256, Unsigned},
 };
-use module_lattice::{Field, Truncate};
+use module_lattice::{Field, MaybeBox, Truncate};
 
 /// Algorithm 39 `MakeHint`: computes hint bit indicating whether adding `z` to `r` alters the high
 /// bits of `r`.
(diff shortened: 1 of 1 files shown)


tarcieri:
Hmm, `Signature` directly `MaybeBox`es `Vector`. I guess that is to say: should these types always `MaybeBox` themselves, or should we have `Signature` do it, or a mixture...

Edit: `Vector` in particular comes from `module_lattice`, which also defines `MaybeBox`. So it could self-`MaybeBox` if we wanted it to.

baloo:
We can push `Vector` to use `MaybeBox`, but that would not solve the issue cause by `Hint` which does not use `Vector`s in the first place.

Also from a release perspective, there is a combination of version where we're going to end up with double MaybeBox. Probably acceptable.

tarcieri (approved):
I mostly brought that up as a consistency issue.

It seems generally preferable to put things in `MaybeBox` as early as possible, so ideally they can just get allocated directly on the heap when the `alloc` feature is enabled, and don't spend a lot of time on the stack before being moved into a (Maybe)`Box`.

Agree on updating `Vector` similarly and that it may lead to some double-`Box`ing until all crates are updated.

baloo:
we can't make Vector hold a MaybeBox because of the const of Vector::new:
https://github.com/RustCrypto/KEMs/blob/5ab0200cacc1967b15731c6b01cc348cd9e372ca/module-lattice/src/algebra.rs#L260

tarcieri:
oof

